PR #27, opened 2018-04-10
Update O365 ProPlus version #s, update links to O365 ProPlus content

DHB-MSFT:
Also added link break between links in "Next steps" section so they appear on separate lines.

Windows 10 version numbers need to be updated as well, but not sure what correct version #s are.


lizap:
Thanks, Dan. Merging now. I'll reach out separately re: Windows versions.

PR #44, opened 2018-05-16
Update .openpublishing.redirection.json

denisebmsft:
Need to redirect this URL: https://docs.microsoft.com/en-us/microsoft-365/compliance/gdpr-dsr-o365
to this: https://docs.microsoft.com/en-us/microsoft-365/compliance/gdpr-dsr-office365


denisebmsft:
Please review and approve so we can get these redirects working as expected. Thank you!

Issue #134, opened 2018-10-20
Correction required


santoshseth:
prioritize what you to move - should be "prioritize what you want to move" in the first line of this page.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 25e0c748-09b6-1a61-6c01-4fafd1e47160
* Version Independent ID: f0a98175-92fd-ae6e-5bdd-ab1de2435d3d
* Content: [Step 1 - Device and App Readiness](https://docs.microsoft.com/en-us/microsoft-365/enterprise/step-1-device-and-app-readiness#feedback)
* Content Source: [microsoft-365/enterprise/step-1-device-and-app-readiness.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/master/microsoft-365/enterprise/step-1-device-and-app-readiness.md)
* Service: **o365-solutions**
* GitHub Login: @JGruszczyk
* Microsoft Alias: **jogruszc**

matthiaspfenninger:
Created PR #136 for this issue.

JGruszczyk:
Updated spelling and fixed typo + broken links

Issue #132, opened 2018-10-18
Link is not helpful


DeanGross:
The link in the Note to the Modern Desktop Deployment Center just opens a marketing site for Microsoft. The aka.ms address know longer works.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ec00a1bd-2fc6-6275-b4ac-c5f7526de0c7
* Version Independent ID: 8ba861e3-c974-a765-d1ff-e07a864c990a
* Content: [Getting Started - Modern Desktop Deployment](https://docs.microsoft.com/en-us/microsoft-365/enterprise/getting-started)
* Content Source: [microsoft-365/enterprise/getting-started.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/master/microsoft-365/enterprise/getting-started.md)
* Service: **o365-solutions**
* GitHub Login: @JGruszczyk
* Microsoft Alias: **jogruszc**

matthiaspfenninger:
Created PR #136 for this issue.

JGruszczyk:
Addressed 136 but fixed spelling, links

PR #136, opened 2018-10-24
Fix typo and broken link (issues #132 and #134)

matthiaspfenninger:
Fixes the error in the first sentence from issue #134


matthiaspfenninger:
Fixes the broken link reported in issue #132

PR #186, opened 2018-11-28
extra space removed

mertenar:
extra space in parenthesis around Windows Defender removed


lizap:
Thanks for the fix - you'll see the change live today. Thanks!!

PR #206, opened 2018-12-06
Updated MMD Admin contacts article

ElaineAnsell:
Two sections updated: 
1.) RBAC roles: added sentences to describe that these roles have the same permissions on the Admin portal, that their role doesn't restrict them from only doing certain tasks.
2.) Add admin contacts: added sentences to describe why and how Admin contacts will be contacted


jdeckerms:
@ElaineAnsell Thanks, Elaine! I changed the instances of "MMD", we don't use that in customer-facing docs.

ElaineAnsell:
Thank you, Jeanie!

PR #216, opened 2018-12-14
Accuracy refresh update - device configuration

mertenar:
These are primarily accuracy changes from SMEs. Removal of "Advanced policies" section and Exploit guard with it's associated table as these are no longer true to our service. 

Other language revisions made.

### DIFF
--- a/microsoft-365/managed-desktop/service-description/device-policies.md
+++ b/microsoft-365/managed-desktop/service-description/device-policies.md
@@ -21,13 +21,13 @@ Devices will arrive with a signature image, then join the Azure Active Directory
 
 ## Why MDM over Group Policy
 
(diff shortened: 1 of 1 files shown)


jdeckerms:
@mertenar I made a few tweaks before merging. "Mobile device management" isn't a proper noun and isn't capitalized when spelled out. When adding links to microsoft.com pages, always remove the language code (en-us) from the URL so the URL resolves to the reader's selected region/language (when localized). Also, I removed a misplaced "on". :)

mertenar:
Thanks @jdeckerms, I was unsure on MDM as I googled and saw it both ways in places. And will keep the "en-us" in mind moving forward.

PR #217, opened 2018-12-14
ring based terminology

mertenar:
ring based needs changed to group, aligns with updates section changes coming within the next couple days


jdeckerms:
thanks, @mertenar, I just added the hyphen to "group-based"

PR #282, opened 2019-03-04
Typo

Ronald-den-Os:
In the purple box it says optizming instead of optimizing


robmazz:
@JGruszczyk - please review and merge if appropriate, thanks

PR #368, opened 2019-05-23
Update device-policies.md

mertenar:
removed section that is out of date, content copied for team to revise in the future.

Additional policy around Powershell V2 deprecation added


mertenar (commented):
@jaimeo

PR #381, opened 2019-06-25
Removing references to WDAC & AppLocker

ChristopherTulip:
I checked with Legal and there shouldn't be an issue removing these. We will want to notify LBG & NZTE that we will continue to support them.


mypil:
@jaimeo - PR has been copy edited and is ready for final review. 

Thank you.

PR #410, opened 2019-07-22
Update device-policies.md

mertenar:
removing the App allow list content as it has been removed from the service @DaniHalfin @greg-lindsay


mypil:
@jaimeo - PR is ready for final review. Thank you.

PR #431, opened 2019-08-01
Update add-admin-contacts.md

mertenar:
@jaimeo changes reflect new UI on MMD admin portal


jaimeo:
The "areas of focus" are concepts, right? That is, not formally named entities. If so, then do not use title case (i.e., "IT help desk" and "App packaging").

mypil:
@mertenar - kindly check on the response from jaimeo and apply the suggested changes.

Thank you.

mertenar:
edited per @jaimeo 's suggestion - let's await reply from Stuart (Jaime see thread 8/5) before committing

PR #454, opened 2019-08-28
Update register-devices-self.md

mertenar:
@jaimeo removed error that could result from removing a device


mypil:
@jaimeo  - PR has been copy edited and is ready for final review.

Thank you.

PR #526, opened 2019-10-04
Update regions-languages.md

mertenar:
@jaimeo now transacting in Benelux !


jaimeo:
Let me just check if we can abbreviate like that or have to spell out the names....

mertenar:
Thanks, I was going that it was a region, but yes should make sure we are compliant

PR #567, opened 2019-10-16
Update find-and-release-quarantined-messages-as-an-administrator.md

ryancairnes:
Making changes to reflect that the SCC should be used instead of the EAC and to reflect the change in the retention period.


mypil:
@nenonix @JohanFreelancer9 - kindly do the copy editor review for this PR. Thanks.

PR #572, opened 2019-10-17
Changing a script to avoid confusion

jwmurraycmu:


mertenar:
@jaimeo change to script from James to prevent possible confusion by user

PR #605, opened 2019-10-23
Update policy tip scenarios

shelri:
This change is to place the policy tip working scenarios in line with how the service actually performs. We have received customer escalations due to the lack of clarity, in particular with regard to using predicates with exceptions.


chrfox (commented):
this is a duplicate of #905 which was addressed and closed on 11/5 by the requestor.  Reject this PR.

chrfox:
this is a duplicate of #905, which is already closed.

PR #651, opened 2019-10-30
Updated availability table

MikePaer-MSFT:
Outlook Mobile shipped to 100% GA WW, so updating table to reflect that. 
Also updated the "default label" scenario in the availability table for web - this scenario is not coming soon, it's TBD.


mypil:
@greg-lindsay - PR is ready for final review. 


Thank you.

cc: @kenwith

greg-lindsay:
The conflict was caused by the original table being split into two tables, but the edit here was done on the old single table.  However, looking at the public version https://docs.microsoft.com/en-us/microsoft-365/compliance/sensitivity-labels-office-apps the changes in this PR already appear to be there, so I'll just close this PR.

PR #790, opened 2019-11-20
Update verbage on autoexpanding archives.

lindabrGithub:
@markjjo We need to update text on Unlimited archives not really be unlimited.  We have gotten feedback that the AuxArchive are limited to 1 TB of storage.  Can we update this document to state:> Office 365 auto-expanding archiving feature provides up to 1 TB of additional storage in archive mailboxes. 

We need to get this updated in other articles that have this messaging as well.


markjjo:
@lindabrGithub merged

PR #797, opened 2019-11-21
Update SPO AV documentation

swannman:
- Remove "this feature requires ATP" because that is not accurate
- Clarify that AV scanning may be performed on a subset of files based on a heuristic


swannman:
@MSFTTracyP adding you here per the email thread I started. Thanks!

MSFTTracyP (approved):
Thanks

PR #807, opened 2019-11-21
The min specs for the Surface models have been updated

rekhanr:
Surface Book 2 - Intel Core i5, 256GB, 8GB RAM device.
Surface Laptop 2 - Changed from 256GB to 128 GB.
Surface Pro 6 - changed from 256 GB to 128GB.
This matches the specs that are defined in the link associated with each model.


rekhanr:
@mertenar @jaimeo please review.

mypil:
@jaimeo - PR is ready for final review.  Thank you.

PR #813, opened 2019-11-22
Adding AIR events in Management API

vippand:
Adding description about AIR events in Management API


denisebmsft (approved):
@vippand looks good! Thank you!

PR #914, opened 2019-12-10
Update to clarify for new disposition feature

robertoyms:
Clarification of disposed items including items which are also already disposed and records.


markjjo:
@robertoyms merged PR

PR #917, opened 2019-12-10
Update how-atp-safe-links-works.md

ruiwensu:
Changed from "on the web" to "in the browser" because there could be other thick clients with embedded WXP capabilities that connect to the web that isn't supported, so "in the browser" is a more accurate description.


MSFTTracyP (commented):
Thanks

MSFTTracyP (approved):

PR #959, opened 2019-12-18
Added another example to AttachmentNames

mahage-msft:
@markjjo - Please review and publish


markjjo:
@mahage-msft PR merged

PR #990, opened 2019-12-28
Changed link from Trust Center > GDPR to the homepage of Trust Center

GiantCrocodile:
I think this link fits better as the topic isn't primary about GDPR.


mypil:
@BrendaCarter - PR is ready for final review.  

Thank you.

cc: @kenwith

robmazz:
@GiantCrocodile - thanks for the feedback. Closing this suggestion without merging, the link in question was removed as stale.

PR #1000, opened 2020-01-02
Updating records page to clarify licensing

robertoyms:
We've realized that it wasn't clear from the documentation that declaring an item as a record with retention labels requires E5-level licensing. Adding a note to clarify this.


mypil:

JohanFreelancer9 (approved):

mypil:
@cabailey - PR has been copy edited and is ready for final review. 

Thank you.

cc: @kenwith

cabailey:
@robertoyms  Thanks for making this clarification!  To avoid ambiguity with other licenses that might be required to edit content, should be it "don't require this license"?

robertoyms:
Yes that works.

Roberto.

cabailey:
@robertoyms  I don't see the additional clarifying words added yet, so for expediency, I'll merge this & do another PR myself to complete the sentence so it can live today.

PR #1058, opened 2020-01-13
Update use-dkim-to-validate-outbound-email.md

andypunt:
removing 1028 key migration reference


mypil:
@MSFTTracyP - PR is ready for final review.  

Thank you.

cc: @kenwith

PR #1064, opened 2020-01-14
Update ediscovery-cases.md

lindabrGithub:
@markjjo 
Can we add the note about deleting a case.  Support has been getting few cases on this where a customer tries to delete a case ; however it still have  a hold, even though the hold is off are turned off the case can't be deleted.   Sometimes customer wants to delete the case and but the hold must be deleted before the case can be deleted.


markjjo:
@lindabrGithub I went ahead and added a new step on deleting cases and included the warning about deleting holds before you can delete a case: https://docs.microsoft.com/en-us/microsoft-365/compliance/ediscovery-cases#optional-step-11-delete-a-case

Closing this PR

PR #1065, opened 2020-01-14
Update create-a-custom-sensitive-information-type-in-scc-powershell.md

lindabrGithub:
@chrfox  Can we add -ReadCount 0 , after -Encoding Byte) in the [New-DlpSensitiveInformationTypeRulePackage] example here.  If not, you will receive an error regarding the object size is not allowed.  This is shown in the Powershell article for [New-DlpSensitiveInformationTypeRulePackage].

Thanks


chrfox:
merged

PR #1128, opened 2020-01-24
Removed a reference to Search-Mailbox

mahage-msft:
@markjjo - Please review and publish.


markjjo:
@mahage-msft I did not merge this PR because you removed the bullet about the 10 item limit, which is an important point. You can remove reference to Search-Mailbox, but I don't think we should remove 10 item limit

PR #1137, opened 2020-01-25
improved readability of paths & file names by using inline code lines

GiantCrocodile:


chrfox:
Hi @GiantCrocodile 
Thank you for your suggestions. I've worked in some emphasis for readability as well as other changes to the topic

PR #1184, opened 2020-02-03
Update use-dkim-to-validate-outbound-email.md

andypunt:
Update DKIM config to account for possible missing selector.

### DIFF
--- a/microsoft-365/security/office-365-security/use-dkim-to-validate-outbound-email.md
+++ b/microsoft-365/security/office-365-security/use-dkim-to-validate-outbound-email.md
@@ -113,7 +113,7 @@ To configure DKIM, you will complete these steps:
 
 For each domain for which you want to add a DKIM signature in DNS, you need to publish two CNAME records.
 
-Run the following commands:
+Run the following commands to create the selector records:
 
 ```powershell
     New-DkimSigningConfig -DomainName <domain> -Enabled $false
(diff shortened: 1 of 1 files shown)


mypil:

microsoft-365/security/office-365-security/use-dkim-to-validate-outbound-email.md
```diff
@@ -177,6 +175,9 @@ Points to address or value:	selector2-cohowinery-com._domainkey.cohovineyardandw
 TTL:				3600
 ```
 
+> [!NOTE]
+> Creating the second record is important to do but only one of the selectors may be available at the time of creation. In essence, the second selector may point to an address that has not been created yet. It is still recommended that the second CNAME record is created since it will come in handy when the keys rotate and, as an admin, there would be nothing to do and the rotation would be seamless.
```
JohanFreelancer9:
```suggestion
> It is important to create the second record but only one of the selectors may be available at the time of creation. In essence, the second selector may point to an address that has not been created yet. It is still recommended that the second CNAME record is created since it will come in handy when the keys rotate and, as an admin, there would be nothing more to do and the rotation would be seamless.
```

JohanFreelancer9 (changes requested):
Copy review - suggestions only.

mypil:
@MSFTTracyP - PR has been copy edited and is ready for final review. 


Thank you.

cc: @kenwith

PR #1187, opened 2020-02-04
Threat trackers wrong description

Tosue:
As the description in https://protection.office.com/threattracker says, "The trending campaigns display dynamic assessments of email threats received your Office 365 service. It is designed to show you trending on received malware, so that you know which threats are increasing and need your focus." So the article is wrong, it's trending malware Received, not that haven't been seen.


denisebmsft (approved):
Thank you @Tosue

PR #1198, opened 2020-02-05
Update status state for managing jobs-ediscovery20.md

telaws:
We have had a few escalations where customers would like to know why they see a status of "Partially Successful" when performing an advanced eDiscovery search on custodian location.  However, it only indicates description for "Partially Successful" = "The job was partially successful."  When it actually means The job did not find any unindexed data in any of the custodian locations and the job is partially successful.  Would like documentation to be updated to include that in order to avoid more escalations going forward.


mypil:

JohanFreelancer9 (approved):

mypil:
@markjjo - PR has been copy edited and is ready for final review. 


Thank you.

cc: @kenwith

markjjo:
@telaws Thanks for submitting. I tweaked your description: "This status is typically returned when the job didn't find any partially indexed data (also called *unindexed data*) in some of the custodian data sources."  Let me (markjjo) know if you don't like this change.

PR #1293, opened 2020-02-26
Update device-list.md

rekhanr:
Removed * for two HP models as newer models are added.
Added * to the new DELL models published in the last change list
Added HP EliteBook x360 1040 G6
Moved HP EliteBook x360 1040 G5 to archived, as this is also affected by the intel chip shortage.


rekhanr:
@jaimeo , @jwmurraycmu - this is the chnagelist to add the new HP model.

PR #1379, opened 2020-03-06
Removed "that's available"

jaimiefox:
Removed "that's available" in line 93.


markjjo:
@jaimiefox PR is merged. Your contributorship should be live by EOD

PR #1407, opened 2020-03-11
Fixed typo which may lead to confusion

GiantCrocodile:
subdomain (`profile` vs `profiles`) didn't match to previous example


MSFTTracyP (commented):
Adding

PR #1409, opened 2020-03-11
Update customer-key-availability-key-roll.md

brdesai:
Swapped the order for availability key vs. Customer Key rolling.


mypil:
@KCCross - PR is ready for final review.  

Thank you.

cc: @kenwith

PR #1436, opened 2020-03-16
Update microsoft-secure-score-whats-coming.md

B3Jester:
@levinec what do you think about handling TVM like this rather than listing out the massive list of controls?  is there somewhere in the TVM docs we can link to for a list?
Can we do another level of header?

Let's not publish until Thursday am either way, as i've given PMs until eod wednesday for any changes to this list


chris24p:
It's all in the set of origins the first two or three words make up the whole entire system. Plus it won't work without the two functioning as a number separated by comas.also publishing requires a bit of unification listed on a separate Network that just ain't the biggest in the universe

B3Jester:
@chris24p I’m not sure what your comment means in this context... can you clarify?
@levinec for awareness

PR #1444, opened 2020-03-18
Fixing link in datastore and rulepack example xml.

wjwmichael:
Fixing link in datastore and rulepack example xml. Should be http instead of https in both examples.


chrfox:
accepted and merged

PR #1445, opened 2020-03-18
Added SearchMailboxes operation

mahage-msft:
@markjjo - for your review.


markjjo:
@mahage-msft PR merged; should be live later this morning or this afternoon

PR #1461, opened 2020-03-21
Removed unsupported scenario.

ericspli:
Remove the following comment as iPad is NOT supported for Centralized Deployment
o	  On iOS (iPad only), Version 2.9.18010804 or later.


mypil:
@Sirkkuw - PR is ready for final review.  

Thank you.

cc: @kenwith

Sirkkuw:
This has already been fixed in the live topic.

PR #1470, opened 2020-03-23
Note about Keywords field

mahage-msft:
@markjjo - The Keywords field can include any of the fields listed in the Searchable field name using KQL, this is similar to how workload search works, except in AeD there are lots more conditions so people don't use this approach often.  Feel free to re-word or give me a shout if you have any questions.


markjjo:
@mahage-msft PR editing and merged

PR #1482, opened 2020-03-24
updated new SKU for AME

samschan-ms:


samschan-ms:
@KCCross  please review the change of SKU for AME.

samschan-ms:
@KCCross  You need to hold this publish until 4/1

KCCross:
Published 4/1.

PR #1543, opened 2020-04-01
Added missing quote mark

mahage-msft:
@markjjo - for your review


markjjo:
@mahage-msft PR merged

PR #1592, opened 2020-04-08
Enhance short meta descriptions

walsehgal:
Enhance short meta descriptions to be more relevant and improve SEO.


walsehgal:
@DaniHalfin

PR #1596, opened 2020-04-09
Adding note about duplicate records

get-itips:
https://github.com/MicrosoftDocs/office-docs-powershell/issues/4677


get-itips:
@chrisda Chris, please review if you agree with adding this note.

PR #1608, opened 2020-04-11
Update switch-to-a-different-plan.md

bpeterse03:
Think we should remove the line to call support. Would like to see the customer attempt switching plans manually before calling support.


mypil:
@cmcatee-MSFT - PR is ready for final review. 

Thank you.

cc: @kenwith

cmcatee-MSFT:
This article will be deprecated for WW customers soon, no change needed at this time.

Issue #1500, opened 2020-03-27
Examples are always great, Step 1 needs one badly


ITMHelps:
A sample command in Step 1 would go miles to making this article useful.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 464a4f5f-8740-79aa-3647-a9b5e351656c
* Version Independent ID: a51103db-92b2-d225-d493-81d624e6fc6b
* Content: [Search for and delete email messages in your Office 365 organization - Microsoft 365 Compliance](https://docs.microsoft.com/en-us/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization?view=o365-worldwide#feedback)
* Content Source: [microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md)
* Service: **o365-seccomp**
* GitHub Login: @markjjo
* Microsoft Alias: **markjjo**

get-itips:
Hello @ITMHelps how are you?
Created a PR to propose adding a working example to step # 1.
If it gets approval, it will be reflected on the article.
Thank you

mypil:
@ITMHelps - Thank you for submitting feedback.

From our understanding, the issue has been resolved based on this merged commit https://github.com/MicrosoftDocs/microsoft-365-docs/commit/9d4cf3991dcca2d04733b86248e5d8014a657d2f. 


Thank you for your contribution to make the docs better! Much appreciated!

PR #1625, opened 2020-04-14
Adding an example

get-itips:
Adding an example to the Step # 1

Also, corrected the current use of the -SearchName parameter for the New-ComplianceSearchAction cmdlet which does not accepts the search name using ""

Reference Issue https://github.com/MicrosoftDocs/microsoft-365-docs/issues/1500

### DIFF
--- a/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md
+++ b/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md
@@ -89,11 +89,20 @@ Here are two examples of queries to find suspicious email messages.
     (From:chatsuwloginsset12345@outlook.com) AND (Subject:"Update your account information")
     ```
 
+Here is an example use of the search query by running the **New-ComplianceSearch** and **Start-ComplianceSearch** cmdlets specifying to search on all mailboxes.
+
+```powershell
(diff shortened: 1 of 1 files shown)


markjjo:
@get-itips Thanks for doing these. However, why you are  adding the additional step to create the variable, particular in the compliancesearchaction examples? The previous examples are technically correct and not sure the advantage of variable? AFAIK, there isn't a standard to do this.

e.g.

```powershell
$SearchName="Remove Phishing Message"
New-ComplianceSearchAction -SearchName $SearchName -Purge -PurgeType SoftDelete
```

get-itips:
> @get-itips Thanks for doing these. However, why you are adding the additional step to create the variable, particular in the compliancesearchaction examples? The previous examples are technically correct and not sure the advantage of variable? AFAIK, there isn't a standard to do this.
> 
> e.g.
> 
> ```powershell
> $SearchName="Remove Phishing Message"
> New-ComplianceSearchAction -SearchName $SearchName -Purge -PurgeType SoftDelete
> ```

Hello Mark, how are you? 👍 

The cmdlet New-ComplianceSearchAction is not currently accepting string literals for the SearchName parameter, I tested this and the only way that it accepted the parameter value was supplying it as a string variable.
It just won't find the Compliance Search:

image

So I thought in providing a working example that people can actually use.

markjjo:
@get-itips I should have also said that creating the variable for the New- and Start-ComplianceSearch examples does make sense and is a good suggestion. Thanks for heads up about error. I will test as well.

PR #1638, opened 2020-04-15
Update meta descriptions

walsehgal:
Shorten and remove majority of reserved characters from meta descriptions.


walsehgal:
@DaniHalfin

PR #1654, opened 2020-04-17
Anchorlink text edits

jdmartinez36:
Edited anchorlink text for links that used "here." Also made minor edits to the surrounding sentence to make sure it made sense after the change.


jdmartinez36:
@DaniHalfin

PR #1701, opened 2020-04-22
Used example external IP address and not an internal address

brianreidc7:
The edit changes the section that says the IP address(es) that you add to the SPF record would look like an internal IP (starting 192 in the original document). The internal IP of the server is not what is needed in the SPF record and it must be the IP address as seen by Exchange Online Protection, aka the external IP address. Added "External" to the the sentence and changed the internal IP example to an example external IP (this IP was used by Microsoft Learning years ago as an example external IP and so is safe to use as is owned by Microsoft.


mypil:

JohanFreelancer9 (approved):

mypil:
@MSFTTracyP  - PR has been copy edited and is ready for final review. 

Thank you.

cc: @kenwith

Issue #1653, opened 2020-04-17
Doc out of date


Punk-Geek:
Does not reflect changes to Office Admin interface. There is no "Security & Privacy" page under settings.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ad408574-36cc-9b96-f092-58a68708d3a0
* Version Independent ID: 97adbc3d-8f71-a6de-25c3-f9400125f885
* Content: [Let users reset their own passwords in Office 365 - Microsoft 365 admin](https://docs.microsoft.com/en-us/microsoft-365/admin/add-users/let-users-reset-passwords?view=o365-worldwide#feedback)
* Content Source: [microsoft-365/admin/add-users/let-users-reset-passwords.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/admin/add-users/let-users-reset-passwords.md)
* Service: **o365-administration**
* GitHub Login: @cmcatee-MSFT
* Microsoft Alias: **cmcatee**

msbemba:
@Punk-Geek Thank you for your feedback. We have Modified the document to reflect the steps via PR #https://github.com/MicrosoftDocs/microsoft-365-docs/pull/1766

Thanks
Sri

mypil:
@Punk-Geek - Thank you for submitting feedback.

From our understanding, the issue has been resolved based on this merged commit https://github.com/MicrosoftDocs/microsoft-365-docs/commit/f72fd36f1ff228a3f99d580e28782c2b8e8b3122. 


Thank you for your contribution to make the docs better! Much appreciated!

PR #1766, opened 2020-04-27
Update let-users-reset-passwords.md

msbemba:
Made changes to the instructions to setup self service password reset.


msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/1653

PR #1900, opened 2020-05-13
Update customizing.md

bmathew17:
Update intro

### DIFF
--- a/microsoft-365/managed-desktop/service-description/customizing.md
+++ b/microsoft-365/managed-desktop/service-description/customizing.md
@@ -10,7 +10,7 @@ ms.collection: M365-modern-desktop
 
 # Exceptions to the service plan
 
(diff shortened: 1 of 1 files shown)


jaimeo:
A problem with the way this is worded is that it basically says that MMD only provides those benefits on devices that stay up to date. But really, staying up to date is one of the benefits itself.... Also, "modern device" is jargon we are trying to phase out.

bmathew17:
Fair comments. Just wish this would align with other messaging, also "pleasant experience" just sounds out of place here

jaimeo:
Let's work this out internally. There's no reason that things have to be verbatim everywhere as long as the meaning is the same.

yogkumgit:
@jaimeo Please merge this PR. Thanks.

jaimeo:
Again, we'll work this out internally. If the contributor is a Microsoft person, please come to my office hours so we can discuss.

PR #1909, opened 2020-05-13
Fix script for localization

adamjbrewer:
The script as provided did not work, as it attempted to run two Set-Label cmdlets that independently set the display name and tooltips. These must be executed in one cmdlet, so I combined them into one Set-Label cmdlet and the script that works as expected.


mypil:
@cabailey - PR is ready for final review. 


Thank you.

cc: @kenwith

cabailey:
@adamjbrewer Thank you for flagging this!  The script used to work but it seems there has been a recent change on the backend that now prevents you from specifying a localized name without a tooltip (and vice versa)  - I suspect, to enforce parity for the same locale. I'll merge this PR and I've also reached out to the folks who own the PowerShell ref doc for Set-Label to see if we can add this requirement that both need to be specified together.

PR #1855, opened 2020-05-07
Consistent use of OneDrive for Business

adrianwells:
Changes to use the same wording throughout the document to be consistent to reference OneDrive for Business

PR #1937, opened 2020-05-15
Update get-access-to-and-back-up-a-former-user-s-data.md

msbemba:
Made  changes from one drive to One drive for business


adrianwells:
@msbemba you may want to see this #1855 as it appears "OneDrive" is now the new name for "OneDrive for Business"

msbemba:
Thank you @adrianwells. I see that now. We may have to revert the PR and leave the doc as it is. Thank you again for your help in making our Documents better.

Thanks
Sri

PR #2097, opened 2020-06-02
Fixed incorrect information about search.

jcreigt:


jcreigt:
For reference, I am the Program Manager on Outlook Desktop that owns search.

markjjo:
@jcreigt PR merged; should be published to live site by tomorrow late AM. Ping me if you need this sooner as I will have to repo this change in the private repo. I can push changes to public repo to live.

PR #2130, opened 2020-06-09
Update to permissions for disposition

bhraghun:
The existing documentation doesn't explicitly call out that a Global Admin user should also be part of the 'Disposition Management' role to see the disposition management in the M365 Compliance Center. This is a big source of confusion and there are customer escalations around this. I am discussing with Roberto on the rationale behind the current design. But, lets have a temporary update to the documentation so that customers are aware.


cabailey:
#sign-off  Thanks for making this update!

Issue #1621, opened 2020-04-14
You need to put something about the amount of time that the change will take


cedwards-telis:
You need to put something about the amount of time that the change will take before it comes into effect since the amount of time is significant, I think it is 24 hours, and after reading the article you might expect it to be instant.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d3a862d8-4223-7bb7-42f1-9842cbeacf87
* Version Independent ID: 2c3c7e9a-a44d-168e-2083-b1527ab3f61c
* Content: [Allow members to send as or send on behalf of a Group - Microsoft 365 admin](https://docs.microsoft.com/en-us/microsoft-365/admin/create-groups/allow-members-to-send-as-or-send-on-behalf-of-group?view=o365-worldwide#feedback)
* Content Source: [microsoft-365/admin/create-groups/allow-members-to-send-as-or-send-on-behalf-of-group.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/admin/create-groups/allow-members-to-send-as-or-send-on-behalf-of-group.md)
* Service: **o365-administration**
* GitHub Login: @MikePlumleyMSFT
* Microsoft Alias: **mikeplum**

andrewpong:
I thought I was doing something wrong — will using PowerShell cut the time needed for permissions to take effect?

cedwards-telis:
@andrewpong  Unfortunately it will not.

msbemba:
@cedwards-telis Thank you for your feedback. This permission isn’t always granted immediately. It could be instant, but it could also take up to 2 hours. We have added a note to the document to reflect the time it could take for the permissions to take effect. tracking PR # https://github.com/MicrosoftDocs/microsoft-365-docs/pull/2213

Thanks
Sri

mypil:
@cedwards-telis @andrewpong - Thank you for submitting feedback.

From our understanding, the issue has been resolved based on this merged commit https://github.com/MicrosoftDocs/microsoft-365-docs/commit/fd8c2f5c4a7c2778e2b89e8e8a27b0e634a3dc47. 


Thank you for your contribution to make the docs better! Much appreciated!

PR #2213, opened 2020-06-22
Update allow-members-to-send-as-or-send-on-behalf-of-group.md

msbemba:
Added a note that the there could be a delay up to two hours for the permission to take effect


msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/1621

PR #2275, opened 2020-06-29
SEO edits for security design principles

jdmartinez36:
Edited metadata title to better reflect the wording of the trailhead article.

Added description and included keyword.

The SEO expert expressed that the metadata title at least should be changed, but also the intro if possible. Google is ranking the article based on the "sail" in the title and body. I left the H1 alone since the subject of the article uses that express terminology, but edited the introduction to remove "sail."

@DaniHalfin


tuco-ulf:
@DaniHalfin @BrendaCarter review please

BrendaCarter:
Please keep "sailing over these hurdles" in the text because it matches the title and theme (instead of "overcoming these obstacles.")

PR #2321, opened 2020-07-06

SeunginLyu:
Remove wrong '/' within a query example
@markjjo for your review


markjjo:
@SeunginLyu PR merged. Thanks for catching this.

PR #2322, opened 2020-07-06
Private channel retention for TEams

anwara000:
Removed part of the article that suggested retention for private channel works. Retention is not supported for Private channels.


anwara000:
@MicrosoftHeidi  FYI

cabailey:
Closing - fixing with #4235

PR #2323, opened 2020-07-06
Retention delete latency

anwara000:
Retention on Teams takes seven days to delete content, not three. The same has been updated in Teams documentation

https://docs.microsoft.com/en-us/microsoftteams/retention-policies


anwara000:
@MicrosoftHeidi  FYI

PR #2325, opened 2020-07-07
Update importing-pst-files-to-office-365.md

adramasu:
Updated the Network upload availability regions @markjjo

### DIFF
--- a/microsoft-365/compliance/importing-pst-files-to-office-365.md
+++ b/microsoft-365/compliance/importing-pst-files-to-office-365.md
@@ -142,10 +142,9 @@ Additionally, to create import jobs in the Security & Compliance Center, one of 
   
  **Where is network upload available?**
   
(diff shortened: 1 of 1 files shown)


markjjo:
@adramasu Thanks for making this change. But I see that you also deleted the note that states the PST import is not available in Switzerland. You added Germany to the list, but should we retain the note and state that it still isn't available in Switzerland?

adramasu:
> @adramasu Thanks for making this change. But I see that you also deleted the note that states the PST import is not available in Switzerland. You added Germany to the list, but should we retain the note and state that it still isn't available in Switzerland?

Hi Mark,

I did that deliberately so that we maintain the same standard for other regions for which we dont have support ex: Norway, South Africa to name a few. We are actively working on the new regions and as and when we get this done, we can update the regions supported in the list/

markjjo:
@adramasu PR merged; thanks for response to my question

PR #2330, opened 2020-07-07
Updated EBCM Service Resiliency Guide

tabriggs:
@chrfox - you updated the images (thanks for the help) and we have updated the content. Thanks for working with us and let us know when they should be live! @chrfox


chrfox:
Changing name of topic may break inbound links, closing this PR without merging

PR #2344, opened 2020-07-10
Update classifier-getting-started-with.md

clin16:
Adding a note about gov cloud availability per a PM's question. Trainable Classifiers are in the process of deploying to gov clouds but likely will not complete until August/September timeframe.


chrfox:
@clin16

 Closing this unpublished. We never make promises in our content about future releases as they usually change and MSFT can be held legally liable for statements like this.  This type of information belongs on the product roadmap in the admin center.

clin16:
@chrfox Can we tweak it to say ,"Trainable Classifiers are currently not available in government clouds" then?

PR #2351, opened 2020-07-13
Update import-hr-data.md

adramasu:
@markjjo Please look into this PR. Basically, we have different scripts in GitHub based on the environment that the user is in (public cloud, gov cloud). So, we want the user to use the link to the script in the product rather than having separate links for each cloud in the documentation. Hence this change.


markjjo:
@adramasu Not sure I agree with this: Why don't you want links in documentation? What happens if user doesn't keep tab with Github links open? Also, the first change in file doesn't make sense to me; what does "finish step" refer to? Can you please send me flight info/credentials so that I can test/repo this new workflow?

markjjo:
@adramasu I've looked into this some more using a test tenant. I think I better understand your changes, though I still don't understand the first change in the doc: "In the finish step of Go to [https://compliance.microsoft.com](https://compliance.microsoft.com) and then click **Data connectors** in the left nav." Not sure what "In the finish step" means. 

I assume that you want the user to click on the link in the "Your connector was created" window and then remove if from the topic? If so, I will do that. 

image

PR #2352, opened 2020-07-13
Update for Compliance centre

fredericlapierre:
@cabailey, I have made the change for the update concerning the compliance centre
fred


cabailey:
@fredericlapierre Thanks!  As discussed, will let this go through standard processes.

cabailey:
@sign-off

PR #2378, opened 2020-07-17
Update office-365-groups.md

nickadam:
Calendars are not supported. See FAQ, https://support.microsoft.com/en-us/office/calendar-sharing-in-microsoft-365-b576ecc3-0945-4d75-85f1-5efafb8a37b4


MikePlumleyMSFT:
@nickadam the article refers to the shared calendar that is included as part of a Microsoft 365 group, not the sharing of individual calendars with groups.

PR #2399, opened 2020-07-21
Update prepare-tls-1.2-in-office-365.md

WorkShay:
We need to alter this article to reflect the content shown in other articles for consistency.


WorkShay:
We need to alter this article to reflect the content shown in other articles for consistency.

PR #2431, opened 2020-07-24
Replaced references to Indexing with Hashing

E-Saggese:
Indexing is a step that occurs within the service *after* upload of the hashed data, the step that is executed prior to the upload is the hashing step (this also makes the terminology consistent with the parameters and messages in the tool).


E-Saggese (commented):
Please double-check if the change in the title "Index and upload the sensitive data" affected the direct link to that section as I don't know if Github automatically updates in-page links to sections.

PR #2508, opened 2020-07-30
Update disposition.md

bhraghun:
This update is to let know customers that we have rolled out the proof of disposal change only to SharePoint and ODB for now.


cabailey:
#sign-off

PR #2541, opened 2020-08-04
SEO update to deploy-threat-protection

tuco-ulf:
@DaniHalfin 
"network security" and "IT network security" are popular high value keywords used in searches for this subject, and are found in competitor's pages. Currently "network" is used 0 times in this doc, so I placed it in various places as described below. Author: if you disagree with these keyword placements please offer suggestions on where they might fit logically within the doc. Other valuable keywords to consider include  "security for network" and "networking security". 
added "network security" to title tag
added "IT network security" to description
added "IT network security" and "network" within content


denisebmsft (approved):
Thank you @DaniHalfin and @tuco-ulf for your great work on this!

PR #2579, opened 2020-08-10
Update communication-compliance-feature-reference.md

JarbasHorst:
- Fixed typos
- Fixed HTML format


robmazz:
@JarbasHorst - thanks for catching these, they will be corrected in a separate update PR.

PR #2580, opened 2020-08-10
Update communication-compliance-case-study.md

JarbasHorst:
Fixed typo


robmazz:
@JarbasHorst - thanks for catching these, they will be corrected in a separate update PR.

PR #2617, opened 2020-08-14
Update technical-reference-details-about-encryption.md

WorkShay:
We are remove some cipher suites which will be "shadow supported" meaning that they won't be publicly supported, but will be quietly supported


WorkShay (commented):
Added in as going to be deprecated

KCCross:
Yes, I like this change better. Shadow support didn't make sense to me. It's either supported or it's not. Following the same deprecation date as TLS makes perfect sense. Thanks!

PR #2631, opened 2020-08-17
Patch 117

msbemba:
Made changes after talking to KC


KCCross:
Two new fwlinks for nitro and foxit PDF readers.

PR #2633, opened 2020-08-17
Updated to call out other controls

RossAdams:
Questions about remote domains and other forms of controls from customers so making this clearer.


MSFTTracyP (approved):
Thanks

PR #2634, opened 2020-08-17
Update office-365-encryption-risks-and-protections.md

msbemba:
Changed Thales to nCipher


KCCross:
We may be changing these back to Thales in the near future. Accepting these changes for now.

PR #2665, opened 2020-08-20
Updating to reflect 100KB in terns not 100K Terms

tweekerz:


robmazz:
Verified with AyushR and ChristopheF

PR #2707, opened 2020-08-27
Added clarification to the use of /GetSession

E-Saggese:
Clarified that the command gives you information about the progress of each upload post-processing which is something admins normally need to see for troubleshooting


chrfox:
this has already been added with the salt procedures and is already live

PR #2727, opened 2020-09-01
Correction on who can receive email notifications

MadPrince:
Corrected a contradiction in the article. On line 60 it previously stated that "reviewers can be individual users, distribution or security groups, or Microsoft 365 groups". But further down the same page on line 89, there is a note saying that you can only specify users or mail-enabled security groups, and that Microsoft 365 groups are not supported. Through my own testing I found that the note is correct, e.g. you can only specify users or mail-enabled security groups. I changed line 60 in the article to remove the incorrect information (e.g. removed "Microsoft 365 groups" and changed "distribution or security groups" to say "mail-enabled security groups").

### DIFF
--- a/microsoft-365/compliance/disposition.md
+++ b/microsoft-365/compliance/disposition.md
@@ -57,7 +57,7 @@ When content reaches the end of its retention period, there are several reasons 
 
 When a disposition review is triggered at the end of the retention period:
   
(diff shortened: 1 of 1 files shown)


yogkumgit:
@cabailey Please merge this PR. Thanks.

cabailey:
@MadPrince  Thanks for flagging this inconsistency.  Yes, the Note has the correct (updated) information & missed the previous statement.
#sign-off

PR #2736, opened 2020-09-01
Update prepare-tls-1.2-in-office-365.md

arindamthokder:
Added About SMTP Non TLS communication


WorkShay (changes requested):
This shouldn't be on this page which is about preparing for TLS 1.2 and compatibility requirements. https://docs.microsoft.com/en-us/microsoft-365/compliance/exchange-online-uses-tls-to-secure-email-connections?view=o365-worldwide is a better site for this information. If rewritten to say SMTP and TLS 1.2 will be the only connections supported after deprecation in our inbound mail flow, that should get moved to the section regarding TLS Support notes with a rewritten title that's more appropriate instead of "TLS 1.2 for Microsoft Teams Rooms and Surface Hub".

tiburd:
@denisebmsft 

Can you please help resolve this PR? Thank you!

traya1:
Relocated the proposed changes in this PR. Closing PR.

PR #2884, opened 2020-09-17
Fix Order of Roles assigned to Role Groups

dbroussa:
In looking at the list of the Roles assigned by default to the Role Groups I noticed that there were some inconsistencies with the listings.  Sometimes DLP was listed before Device and sometimes it was the other way around.

I looked into the correct way to alphabetize lists and found that there were conflicting standards, but the NISO seemed to be the most correct for our purposes.  https://www.niso.org/sites/default/files/2017-08/tr03.pdf

This says that acronyms should be alphabetized by the letters in the acronym as opposed to how the underlying words are spelled.  Section 5 on page 8 for details.

I would also point out that the table of Roles below the listing in the Role Groups lists them in this manner (Device before DLP)


yogkumgit:

PR #2886, opened 2020-09-17
Updated TLS deprecation and specially the behavior for SMTP

arindamthokder:
Many customers ate opening ticket asking what will happen to mailflow without TLS, will it be allowed or not. As we do not have any article to clarify the same, though of adding it here.


KCCross:
Duplicated information. Instead, use a link to the TLS deprecation information in the Technical Reference Details article.

KCCross (changes requested):
This is duplicated information. Instead, link to the single source of truth in the technical reference details about encryption article.

yogkumgit:
@arindamthokder Please look into the suggestion and update accordingly. Thanks

arindamthokder:
I am sorry, but I am not able to find any other link to the TLS deprecation information where it's mentioned that after deprecation, Non-TLS connection will continue to work. Can someone share the link please.

PR #2887, opened 2020-09-17
Update view-mail-flow-reports.md

faoquong:
@siosulli


siosulli:
@faoquong just a quick clarification (for me!) the mailflow status report has this this text:
"It's important to understand that if a message is sent to five recipients we count it as five different messages and not one message."

but this one is the opposite? 
"It's important to understand that if a message is sent to five recipients we count it as one message."  

Thanks
Sinead

faoquong:
@siosulli the goal is to call out the differences they have. that is what I'm trying to achieve.
Just FYI we will be deprecating the "Sent and Received mail" report since it is a old report version.

PR #2928, opened 2020-09-22
Updated link.

Lauragra:


markjjo:
Hi @Lauragra is there a reason why you added the locale to the URL you revised? Won't that generate a build warning? Can it be removed?

Lauragra:
> Hi @Lauragra is there a reason why you added the locale to the URL you revised? Won't that generate a build warning? Can it be removed?

No, I just copied the link I was sent. ;). Updated.

markjjo:
@Lauragra Thanks for updating! I will merge the PR.

PR #2970, opened 2020-09-25
Add limitation to scope of users and only to email

dbroussa:
This approach is best limited to email only and for a very small number of users for a pilot.  This article should specify those constraints so that users do not think that they can use this approach to start their migration to the cloud for all of the workloads and a large number of users.


yogkumgit:

JohanFreelancer9 (commented):
The copy is fine but we need a link to the article referenced before publishing.

dbroussa:
This isn't a reference to an article, but rather field experience that suggests this modification.  Had a client that did this and rolled out MS Teams as an example and then took MS 2 months to clean everything up because they didn't do a hybrid exchange.  This article, in its published form, implies that this can be a complete pilot of M365 and not just of email.

JohanFreelancer9 (approved):

yogkumgit:
Please merge this PR. Thanks.

PR #2996, opened 2020-09-28
Update classifier-get-started-with.md

clin16:
I've noticed that a few customers are using the seed data as part of the test data. It's best to have a separate set of positives rather than re-use the seed data, so I wanted to clarify this in our support docs.


yogkumgit:
@chrfox Please merge this PR. Thanks.

PR #3007, opened 2020-09-29
Update air-view-investigation-results.md

johnengels:
Updates for the new email cluster type for the user compromised investigations.

This is being added to simplify user compromise investigation, since SecOps teams will want to know what the potentially compromised account sent out.  These clusters automate the search for security teams, but do not trigger cleanup currently, as this traffic is generally headed out from the organization


denisebmsft (approved):
Thanks @johnengels

PR #3023, opened 2020-09-29

clin16:
We recently removed the HTML and CSS detections from our Source Code model. @chrfox


chrfox:
merged

PR #3025, opened 2020-09-30
Update alert-policies.md

johnengels:
Addition of new manual investigation alert - found this page after yesterday's edits.

@denisebmsft - looks like this page has references to the OATP P2 licensing still....


denisebmsft:
@markjjo John Engels added an item (line 111) about an alert that is created when a security operator starts an investigation manually. I've reviewed his contribution. You okay to merge?

PR #3275, opened 2020-10-26
Update office-365-network-mac-perf-overview.md

pandrew1:
Added additional sentence about correlation of test measurement with location as I received a customer question about this


pandrew1:
@kelleyvice-msft please review

kelleyvice-msft (changes requested):
This sentence is a little unclear. I think it should read, "All test measurements from client machines include LAN subnet information, which is correlated with the office location details that you have entered." Does this correctly convey the intended meaning?

PR #3304, opened 2020-10-29
Removed Content Center as the object of a sentence

tomresing:
And updated some capitalization to match the product.


tomresing:
moving to the pubops process

PR #3330, opened 2020-11-02

pamcdo:
I believe the EvoSTS object created by the HCW at one time just named the object 'EvoSTS' so the powershell command to look it up may have worked at one point. That is no longer the case. The object is made with a name of 'EvoSts - GUID' so an 'eq' doesn't work anymore.

I changed this to 'like' with a wildcard on the end and then output to a table to show the full name and the enabled status.

HMA like


pamcdo:
That's not the one I meant to paste in, here's the actual example

hma like 2

pamcdo:
I also noticed, when moving on to the next step in the doc, that the enablement is done by this command

Set-AuthServer -Identity EvoSTS -IsDefaultAuthorizationEndpoint $true

Which isn't going to work via a copy and paste since it's no longer named just EvoSTS. This will probably require a bigger edit to reflect the change that appears to have happened in the HCW.

yogkumgit:

JohanFreelancer9 (approved):

yogkumgit:
@kelleyvice-msft Please merge this Pr. Thanks.

pamcdo:
@kelleyvice-msft bump?

pamcdo:
@kelleyvice-msft one more bump?

pamcdo:
@kelleyvice-msft Just checking on this again

LauraWi:
@dstrome  -- Another item requiring review from Exchange perspective. Thanks.

PR #3346, opened 2020-11-03
Added retirement warning

mahage-msft:
@markjjo - For your review.  Actually, you already edited this in my OneNote so you probably can just submit.


markjjo:
@mahage-msft PR merged

PR #3351, opened 2020-11-04

pamcdo:
This edit is in combination with PR #3330 - modifying the command lines further down and point out that there is the possibility of having the EvoSts server still named without a GUID, depending on when the HCW was run. Re-running the HCW won't update the AuthServer name so these instructions still need to cover both scenarios.


LauraWi:
@dstrome -- more Exchange modern auth that needs to be reviewed.

PR #3366, opened 2020-11-05

ElaineAnsell:
Included the new assignment capability in MEM that now allows PS scripts to be assigned to All users, All devices or All users and devices.  Until this is incorporate into code to reflect Not ready status, it'll be flagged only under Advisory


ElaineAnsell:

PR #3367, opened 2020-11-05

ElaineAnsell:
Updated Advisory guidance for ESP to reflect that the default policy, if enabled, should not be assigned to all devices but should be assigned instead to specific Azure AD groups


ElaineAnsell:

PR #3404, opened 2020-11-10

pamcdo:
There have been some concerns from customers enabling HMA that these URL's to whitelist are just put in the article with no explanation as to what they cover or why they should be allowed to communicate. I have added language to clarify this and also re-worded to get rid of the word 'whitelist'.

I noticed after i had saved changes that the autodetect service link includes some additional data not needed but don't see how i can modify it. You can remove '


LauraWi:
Looping in @dstrome  for Exchange review.

PR #3411, opened 2020-11-10
Added some more SPO datacenters

mahage-msft:
@markjjo - Can you publish this on or about 11/17/2020?


markjjo:
@mahage-msft I will set a reminder to publish on 11/17 as per your request

PR #3418, opened 2020-11-11
duplicate words

billmcilhargey:
Cleaned out duplicate words Microsoft and services


robmazz:
@computeronix - thanks for the feedback, one is a duplicate, the other is by design. Will incorporate in a separate internal PR currently in progress.

PR #3426, opened 2020-11-11

ElaineAnsell:
Added URLs for 3 admins centers referenced in this article-- Azure portal, MEM and OneDrive admin center.

@jaimeo : is this normally done?  It was a suggestion from the CAT team and it makes sense for admins who may not have these admin portals book marked.  I can see them knowing MEM but not Azure portal or OneDive.


jaimeo:
Nope. No links in headings, for several reasons. The intent is fine, but we can't do it that way. I'll add the pointers in text body instead.

PR #3430, opened 2020-11-12
Added export limits

mahage-msft:
@markjjo - for your review.


markjjo:
@mahage-msft PR merged. After this publishes, please verify my revision and let me know if additional limits have to be cited as shared

PR #3480, opened 2020-11-17
Update whats-new-in-office-365-atp.md

shubhanshijainmicrosoft:
The update is regarding an existing remediation export limits feature. Remediation allowed exporting with a smaller capacity.


MSFTTracyP (commented):
Thanks

PR #3506, opened 2020-11-19
Update view-licensed-and-unlicensed-users-with-microsoft-365-powershe…

kenwmitchell:
…ll.md

I think that in the example code for getting licensed users, it should be `$licensed = $False` in the declaration instead of `$licensed = $True`. Otherwise there is never an exception case.

Currently the list returns users with or without assigned licenses. In the submitted PR, it returns a list only of users with assigned licenses.


yogkumgit:

JohanFreelancer9 (approved):

PR #3514, opened 2020-11-19
Added 100K export limit which was referenced in the note

mahage-msft:
@markjjo for your review.


markjjo:
@mahage-msft PR merged

PR #3524, opened 2020-11-20
Updated note on First Contact safety tip

urgandhi:


MSFTTracyP (changes requested):
Let's stick to MarkDown!

Even when the impersonation protection, or impersonation safety tips are turned off, it is recommended you use a mail flow rule (also known as a transport rule) to add a message header named **X-MS-Exchange-EnableFirstContactSafetyTip** with value **Enable** to the messages. Specific safety tips will be displayed notifying recipients that they often don't get email from the sender or in cases when the recipient gets an email for the first time from the sender.

urgandhi:
> Let's stick to MarkDown!
> 
> Even when the impersonation protection, or impersonation safety tips are turned off, it is recommended you use a mail flow rule (also known as a transport rule) to add a message header named **X-MS-Exchange-EnableFirstContactSafetyTip** with value **Enable** to the messages. Specific safety tips will be displayed notifying recipients that they often don't get email from the sender or in cases when the recipient gets an email for the first time from the sender.

Pushed a new commit with the updated text and it's still important to show this as Note so that this gets highlighted on the page as opposed to showing as plain text!

MSFTTracyP (approved):

PR #3538, opened 2020-11-23
Update secure-by-default.md

andypunt:
Looking to add more color to the reason as to why we don't want to honor some allows.


yogkumgit:

microsoft-365/security/office-365-security/secure-by-default.md
```diff
@@ -47,6 +47,9 @@ More information on these overrides can be found in [Create safe sender lists](c
...
 
+## More on why we are doing this
+The spirit of being Secure by default is that we are taking the action that you would be taking if you knew that this was malicious even if there was an allow in place. This is the same action we have been taking on malware for a long time and we moved high confidence phish in parity with that. We have data to support that the FP rate is very low here and if an FP was to happen, then it could be resolved with an admin submission. We also have the data to support that the tenant anti-spam policy and the user Safe sender allows were too broad and causing more harm than good. Another way to put it is that as a security service, we are taking the steps for you to prevent your users from getting compromised. In addition, this is not a full reject in that the items would take the action that is set in your anti-spam policy, which is either Quarantine, Move to Junk folder, or Redirect to an email address.
```
JohanFreelancer9:
```suggestion
The spirit of being secure by default is that we are taking the action that you would be taking if you knew that this was malicious even if there was an allow in place. This is the same action we have been taking on malware for a long time and we moved high confidence phish in parity with that. We have data to support that the false positive rate is very low here and if a false positive was to happen, then it could be resolved with an admin submission. We also have the data to support that the tenant anti-spam policy and the user safe sender allows were too broad and causing more harm than good. Another way to put it is that as a security service, we are taking the steps for you to prevent your users from getting compromised. In addition, this is not a full reject in that the items would take the action that is set in your anti-spam policy, which is either Quarantine, Move to Junk folder, or Redirect to an email address.
```

JohanFreelancer9 (changes requested):
Copy review.

JohanFreelancer9 (approved):

PR #3657, opened 2020-12-07
Update investigating-partially-indexed-items-in-ediscovery.md

Sokarpi:
if a partially indexed item has metadata (metadata is indexed by definition) that matched the search condition(date range) then it will be returned with the regular Indexed items.

See https://docs.microsoft.com/en-us/microsoft-365/compliance/partially-indexed-items-in-content-search?view=o365-worldwide#messages-and-documents-with-partially-indexed-file-types-can-be-returned-in-search-results  👍 
"Similarly, messages with partially indexed file attachments and documents of a partially indexed file type are included in search results when other message or document properties, which are indexed and searchable, meet the search criteria. Message properties that are indexed for search include sent and received dates, sender and recipient, the file name of an attachment, and text in the message body. Document properties indexed for search include created and modified dates. So even though a message attachment may be a partially indexed item, the message will be included in the regular search results if the value of other message or document properties matches the search criteria."


markjjo:
@Sokarpi PR merged. Thanks for making this update.

PR #3675, opened 2020-12-09

clin16:
Added a section in ###Custom Classifiers to include some of the sample categories customers can create for additional guidance. Was going to include in our GA announcement blog but marketing recommended that this be put into support docs.


clin16:
Thanks so much, Christopher!
________________________________

PR #3723, opened 2020-12-15
Update archive-instant-bloomberg-data.md

adramasu:
@markjjo A small note added


markjjo:
@adramasu PR merged. I tweaked the note and then formatted it as another bullet item in the "Before you begin" section.

PR #3768, opened 2020-12-19

SignorelliDenis:
1 - Changed "FYI" to "Definitions"
2 - Changed "I call" to "you should be familiar with some definitions"
3 - There is no parameter which contains *URL* in OutlookAnywhare, we can match both internal and external using *hostname*
4 - As Microsoft is removing "whitelist" and "blacklist" terms, I updated "whitelist" to "allow network traffic from"


LauraWi:
Looping in @dstrome for review/triage for Exchange.

PR #3839, opened 2021-01-05
Update alert-policies.md

ssaini-pm:
Added new default override alert policies to default alert policies table.


markjjo:
@ssaini-pm PR merged. Thanks for adding these new policies. Note that I rearranged the sequence as the policies are listed in alphabetical order.

PR #3841, opened 2021-01-05
Removed reference to SharePoint eDiscovery Center

mahage-msft:
@markjjo - for your review.


markjjo:
@mahage-msft PR merged. At some point, I need to update the screenshot of the details pane and replace with one from the currently flyout page

PR #3848, opened 2021-01-05
Update office-365-service-encryption.md

brdesai:
Updated EXO info for MMK support.
Reworded text around service encryption benefits to bring clarity.


KCCross:
Excellent update. Thanks, Brijesh.

PR #3858, opened 2021-01-06
Update conditional-access.md

ElaineAnsell:
Added more granular guidance that the Modern Workplace device Azure AD group exclusions are for company created policies, not policies created by MMD and that Modern Workplace service accounts should not be excluded by Modern Workplace created CA policies. @jaimeo , CAT (field team) guidance to add more specificity to avoid customer confusion that could end up making these advised exclusions to MMD created policies.


jaimeo:
Don't merge yet--this needs some rewrite that I'll get to after lunch.

PR #3868, opened 2021-01-07
Clarifying guidance

robertoyms:
Clarifying that the guidance for when a user leaves the org that has been documented assumes the user is a cloud only user (AAD & EXO)


cabailey:
Addressing with internal PR 6450 so gets published more quickly

PR #4000, opened 2021-01-27
Added notes and guidance for the use of the idMatch element to avoid processing overload

E-Saggese:
We are seeing many customers hitting throttling limits (which causes EDM to miss content matches) due to them using Classification SITs that are too broad (e.g. "any number" or "any word"). I added guidance, already reviewed by the PM team, on constraints on the Classification element and recommendations for avoiding this problem.

### DIFF
--- a/microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
+++ b/microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
@@ -204,13 +204,16 @@ In this example, where both `caseInsensitive` and `ignoredDelimiters` are used, 
 
       - **idMatch**: This field points to the primary element for EDM.
         - Matches: Specifies the field to be used in exact lookup. You provide a searchable field name in EDM Schema for the DataStore.
-        - Classification: This field specifies the sensitive type match that triggers EDM lookup. You can provide Name or GUID of an existing built-in or custom classification.
+        - Classification: This field specifies the sensitive type match that triggers EDM lookup. Any time content matching this sensitive type is found in a document or email, that content will be hashed and looked up in the table of sensitive data you provided to see if there's a match. You can provide Name or GUID of an existing built-in or custom sensitive information type for the Classification attribute.
 
       - **Match:** This field points to additional evidence found in proximity of idMatch.
         - Matches: You provide any field name in EDM Schema for DataStore.
       - **Resource:** This section specifies the name and description for sensitive type in multiple locales.
         - idRef: You provide GUID for ExactMatch ID.
         - Name & descriptions: customize as required.
+
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


yogkumgit:

microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
```diff
@@ -204,14 +204,17 @@ In this example, where both `caseInsensitive` and `ignoredDelimiters` are used,
...
...ationally intensive operations and having too many possible candidates passed for lookup can cause the system to reach processing limits and potentially cause actual matches to be ignored. For example, using a regular expression that will match any possible word would cause thousands of lookups for a typical document or email. Adding conditions to the Match element of the EDM sensitive type will not prevent this from happening since that evidence will be evaluated only after the EDM lookup has been performed and a match has been found. You can prevent this by using for the idMatch element a sensitive type that only matches content likely to be the type of information in the EDM sensitive data table, for example by using regular expression that only looks for formatted data or only matches values within the valid ranges, or by including requirements for additional evidence such as the presence of related keywords (as indicated above, adding requirements for additional evidence using the Match element of the EDM sensitive type would not help reduce the number of strings that are evaluated for an EDM match since those conditions would only be evaluated after the lookup has been done).
```
JohanFreelancer9, line 216:
```suggestion
> If you use a custom sensitive information type for the Classification attribute, make sure its definition is narrow enough that it won’t cause a large volume of unnecessary matches. While the EDM lookup will eliminate false positives found by the sensitive information type, hashing and lookup are computationally intensive operations, and having too many possible candidates passed for lookup can cause the system to reach processing limits and potentially cause actual matches to be ignored. For example, using a regular expression that will match any possible word would cause thousands of lookups for a typical document or email. You can prevent this by using for the idMatch element a sensitive type that only matches content likely to be the type of information in the EDM sensitive data table, for example by using a regular expression that only looks for formatted data or only matches values within the valid ranges, or by including requirements for additional evidence such as the presence of related keywords (as indicated above). Adding requirements for additional evidence using the Match element of the EDM sensitive type would not help reduce the number of strings that are evaluated for an EDM match since those conditions would only be evaluated after the lookup has been done. 
```

JohanFreelancer9: @Esaggese - please merge my edits as indicated.

JohanFreelancer9 (changes requested):
Copy review.

yogkumgit:
@Esaggese

microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
E-Saggese: @JohanFreelancer9 I agree with the edits but I just noticed the original text was a bit repetitive on the last point. I suggest removing the following part:

> Adding conditions to the Match element of the EDM sensitive type will not prevent this from happening since that evidence will be evaluated only after the EDM lookup has been performed and a match has been found.


Not sure how to do an edit at this stage of the pull request :-(.

JohanFreelancer9: @Esaggese I have edited my suggestion to include your change (with which I agree). Could you please commit this suggestion? Thanks.

chrfox:
The Note element is much too detailed and much too large.  This also conflicts with the Master version of the file. @Esaggese start an email thread with me on these changes please.

PR #4025, opened 2021-01-30
Encryption for external parties breaks under certain Conditional Acce…

brianreidc7:
…ss rules

I only found a vague reference to this in https://answers.microsoft.com/en-us/msoffice/forum/msoffice_o365admin-mso_exchon-mso_o365b/encrypted-e-mail-error-aadsts90072/e54062b1-1f6d-434d-bd1f-0f58b836de60 and as without this information and that policy in place this does not work and the error is not clear. Therefore added that CA policies need to be considered when sharing encrypted content to external users.

### DIFF
--- a/microsoft-365/compliance/encryption-sensitivity-labels.md
+++ b/microsoft-365/compliance/encryption-sensitivity-labels.md
@@ -300,7 +300,7 @@ Your users type the Gmail email address in the **To** box.  Then, they select th
 
 This label is suitable for sharing very sensitive documents as read-only, and the documents always require an internet connection to view them.
 
-This label is not suitable for emails.
+This label is not suitable for emails and this label requires that the Azure Information Protection app is not blocked or requires MFA/device restrictions in any Conditional Access policy on the source tenant.
 
 1. On the **Encryption** page: For **Assign permissions now or let users decide?** select **Assign permissions now**.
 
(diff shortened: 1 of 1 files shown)


yogkumgit:

microsoft-365/compliance/encryption-sensitivity-labels.md
```diff
@@ -300,7 +300,7 @@ Your users type the Gmail email address in the **To** box.  Then, they select th
 
 This label is suitable for sharing very sensitive documents as read-only, and the documents always require an internet connection to view them.
 
-This label is not suitable for emails.
+This label is not suitable for emails and this label requires that the Azure Information Protection app is not blocked or requires MFA/device restrictions in any Conditional Access policy on the source tenant.
```
JohanFreelancer9, line 303:
```suggestion
This label is not suitable for emails and requires that the Azure Information Protection app is not blocked or requires MFA/device restrictions in any Conditional Access policy on the source tenant.
```

JohanFreelancer9 (changes requested):
Copy review.

yogkumgit:
@brianreidc7

cabailey:
@brianreidc7   Hi Brian, I'm going to put a hold on this for now and look at your feedback more holistically. Where the suggestion is added isn't the only configuration possible that can be used for external access.

But even if we added it in several places, Conditional Access is a different feature that's owned by Azure AD with their own documentation, and similarly for the AIP client. In case you haven't found it, the AIP docs do have some documentation for Conditional Access: https://docs.microsoft.com/en-us/azure/information-protection/faqs#i-see-azure-information-protection-is-listed-as-an-available-cloud-app-for-conditional-accesshow-does-this-work

While I understand why you suggested this addition, it's a risky strategy long-term to attempt to list requirements and dependencies outside our own feature. When you start to list one, customers expect others to be also listed, and an absence could be misinterpreted.

But I do I recognize that Conditional Access is becoming a common implementation that can affect the success of your sensitivity labels when you share content outside your tenant. In fact, guest access is a complex subject all round. I'll have a look at the best way to surface this in our docs, while still directing people to the authoritative information for that feature.

cabailey:
@brianreidc7  Guidance now added here: https://docs.microsoft.com/en-us/microsoft-365/compliance/sensitivity-labels-office-apps?view=o365-worldwide#conditional-access-policies

PR #4030, opened 2021-02-01
Update domains-faq.yml

Ramnay:
The inheritance of the domain in example is incorrect, subdomains www.contoso.com, www.partners.contoso.com till here the inheritance is correct, but for this domain its grand parent domain www.partners.marketing.contoso.com, so it should be www.marketing.partners.contoso.com


yogkumgit:

JohanFreelancer9 (approved):

PR #4048, opened 2021-02-02
Update sp-compatible-pdf-readers-for-irm.md

RedVortex:
Edge is now available for MacOS and is able to open up IRM-protected PDF properly.


RedVortex:
Updated documentation to reflect that Edge for MacOS is now available and able to open PDF that are IRM-protected. I tested Edge on MacOS BigSur using IRM-protected files under Sharepoint and was able to properly open the PDF in the Edge browser.

KCCross:
awesome!

PR #4053, opened 2021-02-03
Update audit-log-retention-policies.md

brucehuy:
We removed the "Create Audit Retention Policy" button on the main page.  So updated the verbiage in step 2 to have them click on the "Audit retention policies" tab.

Need to update the image in step 2 to remove the "Create Audit Retention Policy" button and to move the purple outline around the "Audit retention policies tab" instead.


markjjo:
@brucehuy Thanks for submitting this PR. But I'm going to close it and then make this change and other changes related to the audit retention policies dashboard in the private branch, so that I can run spellcheck and check the formatting.

PR #4064, opened 2021-02-04
Update readiness-assessment-tool.md

ElaineAnsell:
Added that Global Reader admin role is now allowed to run ART. Previously was only Intune Admin and Global Admin. I believe that best practice is to only list the least privileged role (in this case, Global Reader) but here chose to list both Global Reader and Intune Admin as Intune Admin (which is specific to Intune) seems sufficiently different than the Global Reader role (which goes across all admin portals). Is this correct?

@jaimeo , can you please review my edit? Thank you!


yogkumgit:

ElaineAnsell:
Adding @Jaime Ondrusek<mailto:jaimeo@microsoft.com> for review.

JohanFreelancer9 (approved):

jaimeo:
I need to double-check with a colleague on the least privileged role guidance for this scenario.

PR #4092, opened 2021-02-08
Changed deprecate to disable

KCCross:
Added information for SharePoint and OneDrive about required .NET updates.


KCCross:
Added information for SharePoint and OneDrive about required .NET updates.

PR #4183, opened 2021-02-17
Update microsoft-secure-score-whats-new.md

anchawl:
removed graph api from incompatibility.
@levinec


levinec:
@anchawl I feel like we should add something about the new API and how it is now compatible?

PR #4229, opened 2021-02-23
Update m365-apps.md

sh783:
@jaimeo it looks like our docs are out of date. Should be 0 0 3 7, right @ChristopherTulip ?


sh783:
@jaimeo I confirmed that it should indeed be 0 0 3 7. No need for @ChristopherTulip to sign off on this PR.

PR #4309, opened 2021-03-02
Fix PowerShell to update selected sites

marcoscheel:
The sample is making a query to only a few site collections and wants to update/add the label. Set-SPOTenant is the wrong cmdlet. It needs to be Set-SPOSite. Only Set-SPOSite has the "SensitivityLabel" parameter. Set-SPOTenant is only operating at the "farm" level and not at an individual site.


yogkumgit:

cabailey:
@marcoscheel   Are you saying you tried the command as documented and it didn't work?  My understanding is that command as documented on this page is correct, and it's the PowerShell documentation for Set-SPOTenant that needs updating.

cabailey:
The Set-SPOTenant cmdlet has been confirmed by engineering and I've been asked to reject this PR.

PR #4370, opened 2021-03-08
Update admin-support.md

StuEwen:
@jaimeo Updating docs as Customers would not create an incident for that scenario, security incidents can only be declared by security operations center.


Josegto1 (changes requested):
R...

jaimeo:
@Josegto1 what changes are you suggesting? I don't see any other commits or anything....

Josegto1 (approved):

PR #4374, opened 2021-03-08

samschan-ms:
added remove encryption for incoming mail


samschan-ms:
@KCCross please review

PR #4376, opened 2021-03-08
Releasing a policy for retention change

PramodBalusu:


cabailey:
After talking with Roberto, I'm going to address this with an internal PR

cabailey:
Revised section now live, so closing this PR

PR #4389, opened 2021-03-09
Added MCAS endpoint

trashg0blin:


JoeDavies-MSFT:
Updates to endpoints must be done through an internal Microsoft process.

PR #4456, opened 2021-03-17
Update topic-experiences-discovery.md

cjtancortex:
Added a line to expect that suggested topics will not appear instantaneously...not sure if this is the best page to add this.


MikePlumleyMSFT:
Moving this note to [Set up Microsoft Viva Topics](https://docs.microsoft.com/microsoft-365/knowledge/set-up-topic-experiences).

PR #4478, opened 2021-03-19

samschan-ms:
Updated cmdlet check.


samschan-ms:
@KCCross please review

PR #4492, opened 2021-03-22
Update turn-off-directory-synchronization.md

pamcdo:
We have some people asking for a way to convert synced users to cloud-only, turning off directory sync is the way to do it and I have just added a note that disabling synchronization will convert the users to cloud-only.


yogkumgit:

JohanFreelancer9 (approved):

PR #4556, opened 2021-03-29
Move limits from export section over to review set

StefanieBier:
In AeD, there is a distinction between "add to review set" vs final export. The "Add to review set" job lines up with the export limits in Core eD and they need to be added to the Review Set limits. 

The export limits should only include limits related to exporting or downloading from Review set. Please make distinction clear between download and export.


markjjo:
@StefanieBier I revised your suggested changes. Can you please take a quick look and make sure they look ok. Thanks, markjjo.

StefanieBier (commented):
@markjjo Looks good!

markjjo:
@StefanieBier Thanks for reviewing. PR merged.

PR #4615, opened 2021-04-06
Update high-risk-delivery-pool-for-outbound-messages.md

puneethmeister:
This a new feature which is disabled before going world wide, we are modifying the design so this article is currently irrelevant hence deleting the content until we roll out a new design which will be somewhere in may-june'21 timeframe


puneethmeister:
@chrisda can you review the PR

PR #4634, opened 2021-04-07
Update dlp-sensitivity-label-as-condition.md

aamalhot:
Updates for label as a condition GA


aamalhot:
@chrfox please review and publish

PR #4642, opened 2021-04-08
Update linux-install-manually.md

bmrodrigues:
Since this is Linux page we should not have a reference for MacOS. I believe to be a typo


denisebmsft:
Thank you @bmrodrigues

PR #4656, opened 2021-04-09
Corrected inaccurate description of Sensitive by Default

E-Saggese:
Original description says SBD blocks files from being shared until at least one DLP policy scans the new item. This is incorrect, since DLP doesn't do the scanning, and lack of indexing won't prevent DLP from evaluating the document. SBD actually works by blocking access to files until they have been *indexed*, which in turns enables DLP policies to accurately assess the document.


yogkumgit:

PR #4677, opened 2021-04-12
Add SPO Admin Url

heikobischoff:


LauraWi (changes requested):
We can't accept manual changes to office-365-worldwide-endpoints.md   This article is autogenerated from code and manual edits will be overwritten.

heikobischoff:
Reverted changes to office-365-worldwide-endpoints.md, please re-review

PR #4687, opened 2021-04-13
Update configure-exclusions-microsoft-defender-antivirus.md

kurtsarens:
expanding on why exclusions need to be avoided. Defender includes most needed exclusions in the product.


denisebmsft (approved):
thank you @kurtsarens

PR #4699, opened 2021-04-14
Update microsoft-defender-antivirus-in-windows-10.md

TinaMcN:
The localization articles are being machine translated, not human translated as expected.  I think the value is case sensitive.


denisebmsft (approved):
thank you!

PR #4731, opened 2021-04-19
Remove .md for full ref links

f-bader:
Full ref links are broken when .md is used as part of the hyperlink


denisebmsft (approved):
Thank you @f-bader

PR #4840, opened 2021-04-29
Update set-up-devices.md

VinnieFlynt:
@jaimeo , Looking to publish this on 5/3 but need the other doc to go live first that I'm submitting to you so we can reference it in this doc. I'll work with you offline. Please dont publish this before we do that.


jaimeo:
I've absorbed this change into another PR with a new article so that all changes can publish together.

PR #4882, opened 2021-05-03
Update safe-links.md

DanijelkMSFT:
remove row 58 based on feedback from PG Feature PM Brian Wilcox, as this is not the case for Office 365 Apps, instead as soon as the setting is enabling its targeting everyone independent of the safelink policy scope.
-->  But, Safe Links protection for Office 365 apps is only **applied** to users who are included in active Safe Links policies. If a user isn't included in an active Safe Links policy, the user doesn't get Safe Links protection in supported Office 365 apps.


yogkumgit:

chrisda:
@DanijelkMSFT, the removal of that line would cause even more confusion. I Added the line back in as follows:

>  Safe Links protection for Office 365 apps is applied to all users in the organization who are licensed for Defender for Office 365, regardless of whether the users are included in active Safe Links policies or not.

The same statement was also on line 209, so I fixed it there.

This revelation also affects the topic [Configure global settings for Safe Links in Microsoft Defender for Office 365](https://docs.microsoft.com/microsoft-365/security/office-365-security/configure-global-settings-for-safe-links) which I will update accordingly in another PR.

To be clear, it now appears that Safe Links for O365 apps applies to everyone (licensed), but the Block the following URLs list still requires membership in an active Safe Links policy.

PR #4910, opened 2021-05-05
ScanAvgCPULoadFactor

andrewjohnporter:
This is what an Microosft support engineer told me, that manually run scans will ignore this setting and run without any CPU limits


denisebmsft (approved):
Thank you @andrewjohnporter

PR #4918, opened 2021-05-06
Update data-loss-prevention-policies.md

aamalhot:
Minor change as label as a condition is GA


aamalhot:
@chrfox Please review

PR #4924, opened 2021-05-06

shelri:
This is a follow up from a previous PR recommendations on the 4k character search limit


markjjo:
@shelri Thanks for submitting this PR. But I' going to close this and make the change directly in the private branch. This needs editing and other formatting and it's better and less error prone to do it in VSC and private branch. I will send you link to review when it's ready. Target is for next Monday.

shelri:
Thanks Mark

PR #4941, opened 2021-05-07
Update web-content-filtering.md

kurtsarens:
removed section that NP will take over if Smartscreen is disabled as that is incorrect.


denisebmsft (approved):
Thank you @kurtsarens

PR #4985, opened 2021-05-12
Clarification of role groups in ExO & Compliance

wemmer:
Role groups in Exchange Online and Microsoft Compliance are different. Membership in Exchange Online's Organization Management does not give permissions to perform a purge. 

I feel like this is not made clear in the current wording, and can be confusing to Exchange Administrators (including myself) encountering this how-to guide, thinking that they should have the appropriate permissions by already being a member of Organization Management in Exchange Online.

If one is only assigned Organization Management in Exchange Online rather than through Compliance, when importing the PowerShell cmdlets you will not be getting access to the -Purge parameter in New-ComplianceSearchAction, and it will instead throw an error.

### DIFF
--- a/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md
+++ b/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization.md
@@ -36,10 +36,10 @@ You can use the Content Search feature to search for and delete an email message
 
 ## Before you begin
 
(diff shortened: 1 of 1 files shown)


markjjo:
@wemmer PR merged. Thanks for adding this content and thanks for the high-quality documentation!

PR #5003, opened 2021-05-15
Upgrade loc priority metadata value

TinaMcN:
Localized versions are getting very high page views, let's human translate this one.


denisebmsft (approved):
Thank you @TinaMcN

PR #5097, opened 2021-05-23
update-teams-admin-public-preview-process

ChristopherRyanMSFT:
I do not have upload / write access to the image store, so the 2 file paths included in the PR currently do not route to images. The expectation though is that we place the files in that location, matching the specified naming convention. 

Images included here in a .zip for reference:

[teams-public-preview-images.zip](https://github.com/MicrosoftDocs/microsoft-365-docs/files/6528526/teams-public-preview-images.zip)


yogkumgit:

PR #5184, opened 2021-06-01
Update search-for-and-delete-messages-in-your-organization.md

AvrumFeldman:
We cannot create new compliance searches before connecting to o365 so the logical order of the steps should be to first connect and only then create the searches etc...


yogkumgit:

markjjo:
@AvrumFeldman PR merged. Thanks for updating. The existing sequence did support running searches in the GUI and then deleting them in PowerShell. Your change better supports option of running searches in PowerShell.

PR #5207, opened 2021-06-03
Update dlp-sensitivity-label-as-condition.md

aamalhot:
There are some scenarios where Sensitivity label as a condition may not work as expected due to SPO limitations.


aamalhot:
@chrfox please review

PR #5211, opened 2021-06-03
Updating Security center link

melanie-cohen:
Updating from /userTags to /securitysettings/userTags.


melanie-cohen:
FYI @chrisda, proposed update on the link. The former link technically works, but that's a bug that we're working to fix.

PR #5248, opened 2021-06-08
Replace MoCAMP with something understandable

AntoineJo:
MoCAMP is not understandable outside of Microsoft, and even inside.... not always ;)


denisebmsft (approved):
Thank you @AntoineJo

PR #5256, opened 2021-06-08
Update privacy-personal-data.md

sakitong:
true up the existing documentation


sakitong:
update based on the current state

PR #5264, opened 2021-06-09
Update scheduler-setup.md

driesgroblerw:
Updated the link to the "Buy Scheduler for Microsoft 365" page that was broken due to a missing /en-us/ in the link.

PR #5280, opened 2021-06-10
Revert "Update scheduler-setup.md"

JoanneHendrickson:
Reverts MicrosoftDocs/microsoft-365-docs#5264

For localization purposes we do not include EN-US in the link.


driesgroblerw:
But I see the redirect works now. Which it did not before. 👍🏻

PR #5287, opened 2021-06-10
Update office-365-worldwide-endpoints.md

SeanMSFT:
Adding new endpoint being introduced to the service for the SMTP AUTH protocol


LauraWi:
@SeanMSFT Can you contact me via internal email? This topic cannot be manually updated, as it's autogenerated from an internal tool. You need to drive your updates through that process with engineering.

PR #5346, opened 2021-06-16
Upgrade loc priority

TinaMcN:
Loc articles getting high page views, let's get this human translated.


denisebmsft (approved):
Thank you @TinaMcN

PR #5349, opened 2021-06-16

pahuijbr:
@Oludele0315 as discussed changes to doc to reflect UX does not show RTP=On when TP is enabled and customer tries to use GPO.


denisebmsft (approved):
Thank you @pahuijbr

PR #5372, opened 2021-06-21
Fix missing link

AvrumFeldman:
I just googled for the deleted items retention page, there is another one but that is for exchange 2019.


markjjo:
@AvrumFeldman PR merge. Thanks for fixing.

PR #5424, opened 2021-06-24
Update high-risk-delivery-pool-for-outbound-messages.md

puneethmeister:
We are launching a new set of IP pools called Relaypool starting 27th of July and customer communication will start going from end of June. as part of customer comm we also need to update our documentation for new IP pool called as relaypool


puneethmeister:
@MSFTTracyP @chrisda

chrisda:
@puneethmeister, when do you want  this to publish?

puneethmeister:
@chrisda , we can publish it now, MC post for some customers will go live in 3-5 days, so they may search for it

PR #5474, opened 2021-06-28
Update secure-by-default.md

ssaini-pm:
Updating Secure by Default timeline for Exchange mail flow rules from July to August


denisebmsft (approved):
Thank you @ssaini-pm

PR #5541, opened 2021-07-06
Shop Windows 10 Device Link is not working

ChrisSweeneyMSFT:
Current link is dead and needs to be updated to https://www.microsoft.com/en-us/windowsforbusiness/view-all-devices#view-all-filter


jaimeo:
Thanks for catching and fixing that. Just FYI, whenever you include a link to anything at microsoft.com, be sure to remove "en-us" from the URL. This allows visitors to get localized versions of the site based on their browser settings.

PR #5583, opened 2021-07-08
Fixed output so it includes case info

mahage-msft:
@markjjo - for your review.


markjjo:
@mahage-msft PR merged

PR #5593, opened 2021-07-09
added a note on localization to ESR docs

hathin:
Localization experience is inconsistent with ESR enabled, adding a note here for this. ESR is now an optional NOT required setting for MMD, customer need to assess if its right for them.


yogkumgit:

jaimeo:
@JohanFreelancer9 if you're going to do copy edits, it's best to be familiar with the Microsoft Writing Style Guide. We do not use "end user."

PR #5607, opened 2021-07-13
Update archive-ringcentral-data.md

Durga-SC:
The partner link is not active yet, pointing to 404 error hence removed


markjjo:
@Durga-SC PR merged

PR #5704, opened 2021-07-21
Update set-up-anti-phishing-policies.md

urgandhi:
Added exact max limit of 1024 which is what we show on UX anyway in the error dialog box


yogkumgit:

JohanFreelancer9 (approved):

PR #5718, opened 2021-07-22
Making checking Auditing on by default a must

sujitnaray:


sujitnaray:
@markjjo  - Could you please approve this PR?  Customers are complaining about the long tail of Auditing turned on by default, so I want it mandatory for them to check.

markjjo:
@sujitnaray PR merged. Thanks for making this update.

PR #5725, opened 2021-07-23
logic change to the check in ART

hathin:
Reason for this change is that MMD will not automatically fail enrollment if Device Compliance policy is set to All Users. The Admin can choose to assign their own Device compliance policy to MMD users, as long as they are aware that MMD has its own Device compliance policy that will also be applied to MMD users upon MMD enrollment. Thus this check now only has 'Advisory' setting and not 'Not Ready'


hathin:
@jaimeo to review wording

yogkumgit:

JohanFreelancer9 (approved):

yogkumgit:
@jaimeo Please merge this PR. Thanks

jaimeo:
@JohanFreelancer9 if you're going to be doing copy editing, please get familiar with our style guide. We do not use "please" except in certain circumstances (which do not apply in this text). Also, we are starting to move away from "target."

Issue #5734, opened 2021-07-26
Incorrect web link


ceaton001:
The web link for "Shop Windows 10 Pro Business Devices" is invalid. Page does not exist.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3a9af2d6-5aa8-d3be-24cf-56e61a8f5817
* Version Independent ID: 5905b5c1-2e87-205a-7bae-2b5760b5180f
* Content: [Device requirements - Microsoft Managed Desktop](https://docs.microsoft.com/en-us/microsoft-365/managed-desktop/service-description/device-requirements?view=o365-worldwide)
* Content Source: [microsoft-365/managed-desktop/service-description/device-requirements.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/managed-desktop/service-description/device-requirements.md)
* Service: **m365-md**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

msbemba:
@ceaton001 Thank you for your feedback. We have fixed the link via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/5761

Once the author approves this the changes will be live.

Thanks
Sri

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @ceaton001 Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #5761, opened 2021-07-28
Update device-requirements.md

msbemba:
The Site is available only in the US so the URL needs to be https://www.microsoft.com/en-us/windowsforbusiness/view-all-devices

If we remove the locale it results in  404.


msbemba:

msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/5734

JohanFreelancer9 (approved):

yogkumgit:
@jaimeo Please merge this PR. Thanks

yogkumgit:
@jaimeo Gentle reminder

jaimeo:
Looking into why that page isn't following our corporate guidelines re: locales in URLs.

jaimeo:
@msbemba to clarify, it does NOT 404 if your locale is en-us. But it does not appear to be a localized page, so if the locale is other than en-us, it does indeed fail as you report.

PR #5784, opened 2021-07-30
Safelinks for Microsoft Teams released

briwilcox:


briwilcox:
Linking to updated documentation.

PR #5855, opened 2021-08-05
use systemctl edit when modifying service files

### LINKED ISSUE #5540: Incorrect method of configuring static proxy for Microsoft Defender for Endpoint on Linux
This issue relates to an improvement in the documentation to the URL: 
https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/linux-static-proxy-configuration?view=o365-worldwide#post-installation-configuration

To configure the proxy, the documentation asks to set the HTTPS_PROXY environment variable in the Defender for Endpoint service file. Concretely, the documentation ask to edit `/usr/lib/systemd/system/mdatp.service`, but the file should not edited because it can be overwritten by the package manager. It will be better to user "Drop-in Files" function for editing systemd unit as described in the [ArchLinux Wiki Here](https://wiki.archlinux.org/title/Systemd#Drop-in_files).

# Examples:

## Using "systemd edit" command

Run this command:
`sudo systemctl edit mdatp.service`

A file editor like Nano or VI should open. Add your configuration by replacing with your URL.
```
[Service]
Environment="HTTPS_PROXY=http://address:port/"
```

Save the file and restart the service so the changes can be applied.

`sudo systemctl daemon-reload; sudo systemctl restart mdatp`

## By directly editing the file

Create directory:
`sudo mkdir /etc/systemd/system/mdatp.service.d/`

Create file override.conf and add content
`sudo vi /etc/systemd/system/mdatp.service.d/override.conf`

```
[Service]
Environment="HTTPS_PROXY=http://address:port/"
```
Save the file and restart the service so the changes can be applied.

`sudo systemctl daemon-reload; sudo systemctl restart mdatp`

konstruktoid:
This PR:
- Recommends using `systemctl edit --full` in order to create the correct systemd .service file, instead of manually modify the original
- Removes end-of-line spaces
- Closes #5540 

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>


yogkumgit:

JohanFreelancer9 (approved):

PR #5920, opened 2021-08-09
Update search-for-and-delete-messages-in-your-organization.md

Sokarpi:
I am suggesting to add this Note per https://o365exchange.visualstudio.com/Enterprise%20Cloud/_workitems/edit/2261431

Maybe there is a better wording that can be used.
I just want to make sure we cover this topic to prevent confusion.


markjjo:
@Sokarpi Thanks for submitting. I'm going to close this PR because I made the change in the private repo because it was a bit easier to do this in VS code. This change is live. I added an item in the Before you begin section and then I added another note in Step 3. 

https://docs.microsoft.com/en-us/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization?view=o365-worldwide#before-you-begin

https://docs.microsoft.com/en-us/microsoft-365/compliance/search-for-and-delete-messages-in-your-organization?view=o365-worldwide#step-3-delete-the-message

PR #5966, opened 2021-08-11
Update admin-submission.md

arindamthokder:
This has been fixed and we can remove the note of 7 day limitation from the article


arindamthokder:
I think Andy already made an update for the same.

PR #6101, opened 2021-08-19
Update search-the-audit-log-in-security-and-compliance.md

davim1:
Adding newly developed FileRecycled and FolderRecycled audit events. These events are running in production today, but are not yet published within this document.


davim1:
@markjjo i was asked to update docs for new FileRecycled and FolderRecycled events. can you please review?

PR #6121, opened 2021-08-23
Update multi-geo-user-experience.md

elmakhmu:
minor change to extend the list of experiences, where content from different geo-locations can appear in SwM view


elmakhmu:
@MikePlumleyMSFT could you, please, have a loon on this minor change

PR #6131, opened 2021-08-24
Update column names and descriptions

taisuke0710:
Some column names don't correspond with product.

SPO_GroupAccess**ed**ByOwner
SPO_GroupAccess**ed**ByOthers
SPO_TeamAccess**ed**ByOwner
SPO_TeamAccess**ed**ByOthers
* Accessed is correct.

And these metrics are regarding sites, not files.


taisuke0710:
Sorry, I sent wrong PR.

PR #6220, opened 2021-08-31
Update edr-in-block-mode.md

marysiakam9889:
- added benefit for MDAV primary users in first paragraph
- changed MDE to MDAV in the question "should I still use if MDAV is primary" since MDAV is the entity that does the remediation step


denisebmsft (approved):
Thank you @marysiakam9889

PR #6335, opened 2021-09-13
Updated details on /AllowedBadLinesPercentage to explain that default is 0%

E-Saggese:


yogkumgit:

microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
```diff
@@ -444,7 +444,7 @@ This computer must have direct access to your Microsoft 365 tenant.
...
-   If your sensitive information table has some incorrectly formatted values, but you want to import the remaining data while ignoring invalid rows anyway, you can use the /AllowedBadLinesPercentage parameter in the command. The example above specifies a five percent threshold. This means that the tool will hash and upload the sensitive information table even if up to five percent of the rows are invalid. The default value for this setting is one percent. 
+   If your sensitive information table has some incorrectly formatted values, but you want to import the remaining data while ignoring invalid rows anyway, you can use the /AllowedBadLinesPercentage parameter in the command. The example above specifies a five percent threshold. This means that the tool will hash and upload the sensitive information table even if up to five percent of the rows are invalid. In versions of the tool that support this parameter the default threhsold is zero percent so any erroneous lines will cause an error.
```
JohanFreelancer9:
```suggestion
   If your sensitive information table has some incorrectly formatted values, but you want to import the remaining data while ignoring invalid rows anyway, you can use the /AllowedBadLinesPercentage parameter in the command. The example above specifies a five percent threshold. This means that the tool will hash and upload the sensitive information table even if up to five percent of the rows are invalid. In versions of the tool that support this parameter the default threshold is zero percent so any erroneous lines will cause an error. 
```

JohanFreelancer9 (changes requested):
Copy review - @Esaggese, kindly commit the suggested change. Many thanks.

microsoft-365/compliance/create-custom-sensitive-information-types-with-exact-data-match-based-classification.md
E-Saggese: The reason for the proposed wording is that the PM team requested that this is documented clarifying that this applies only to the new version of the tool. This is because the default was different (1%) on the previous versions of the tool (where the value is not configurable) which may continue to be used for months since it is not auto-updated. So either the way I wrote the line or some other wording that indicates that the default is zero only for the version of the tool where the value is configurable would be required. Unfortunately, since the tool does not report a version number, we can't specify it that way, so we have to use the new configurable parameter as the reference.

JohanFreelancer9: @Esaggese - Thank you for the explanation. My suggested change, however, is just to correct the typo ("threshold" instead of "threhsold").

yogkumgit:
@Esaggese Please update. Thanks

JohanFreelancer9 (approved):

yogkumgit:
@chrfox Please merge this PR. Thanks

PR #6512, opened 2021-10-05
updated docs on new functionality

hathin:
current customers can no longer run ART. This tool is just meant to be used for pre-enrolled customers. If a current customer runs the tool, they will always have issues because of configurations we put on their tenant, changed who can run to remove this confusion


yogkumgit:

PR #6544, opened 2021-10-07
Updating Advanced Audit events (E5 events) description

sujitnaray:


sujitnaray:
@markjjo  - Could you review, please? I'm trying to do two things 

1. Make it clear that E5 events are only generated (hence available) after the E5 license is assigned to a user.
2. Moving away from the "crucial events" nomenclature, although I don't have a replacement, so I'm calling them Advanced Audit events.

Also, removed a FAQ which was a bit unclear, and said we published API latencies, which we don't do.

Thanks

markjjo:
@sujitnaray PR merged. Thanks for making these updates. Next time, please submit PRs by first going to the staged version of the topic, which are located at https://review.docs.microsoft.com (just add "review" to the beginning of the URL in the live version). Doing this submits your changes and the PR to the private repo. The advantage of doing this is so I can use spellcheck and other quality control tools which aren't available in the public version. Thanks in advance.

PR #6628, opened 2021-10-18
Update mdatp-macos to version 101.47.27

maximvelichko:
@AirTudor - please review


AirTudor:
I believe we can replace "Hot fix" with something like "Fix for a system freeze occurring on shutdown on macOS Mojave and macOS Catalina". Otherwise we might confuse some customers with internal keywords such as "Hot fix" or "kext"

PR #6675, opened 2021-10-22
Fix MB to GB in calculation description

NewkyNet:


markjjo:
@NewkyNet PR merged. Thanks for fixing this.

PR #6706, opened 2021-10-27
Remove preview from SP/OD integration with AAD B2B

giggio:
SharePoint and OneDrive integration with Azure AD B2B is no longer in preview,
so I updated the text and link.


cabailey:
Great - thank you!

PR #6760, opened 2021-11-02
Update create-and-manage-inactive-mailboxes.md

Smartboss88:
Update doc to make 183 days accurate which starts after hold is removed.


markjjo:
@Smartboss88 PR merged. Thanks very much for making these updates to the inactive mailbox topics.

PR #6924, opened 2021-11-17
Fix path to email notification settings

gh-andrem:


denisebmsft (approved):
Thank you @amueller-tf

PR #6930, opened 2021-11-17
Move note to top of article

lindabrGithub:
@markjjo   Can we move the following note to the top of the article, we are getting escalations to disable this after customers have enabled this for their users and org.  I think if we move the comment up on the article, it would help.

"After you turn on auto-expanding archiving, it can't be turned off. Additionally, administrators can't adjust the storage quota for auto-expanding archiving"


markjjo:
@lindabrGithub It looks like you moved a different bullet to the beginning of article and not "After you turn on auto-expanding archiving, it can't be turned off. Additionally, administrators can't adjust the storage quota for auto-expanding archiving".

I'm going to close this PR and make your suggested change in the private branch. Again, the change will be to move statement that auto-expanding archiving can't be turned off near top of article.

markjjo:
@lindabrGithub Updated article is published. Warning moved as first item in before you begin section.

https://docs.microsoft.com/en-us/microsoft-365/compliance/enable-autoexpanding-archiving?view=o365-worldwide#before-you-enable-auto-expanding-archiving

PR #6942, opened 2021-11-18
Update inactive  mailbox behavior

MSnahuelrey:
The changes in the paragraph is needed as this is not the first support case we have in which clients are confused about the behavior of deleting the hold on an inactive mailbox because of this article line. First part said "If you remove the hold or retention policy from an inactive mailbox and the soft-deleted mailbox retention period for the mailbox has expired, the mailbox will be permanently deleted" and the in the More information section, it is explaining there will be 183 days period retention. Please evaluate to include somehow in that paragraph that will eventually be permanent deleted after 183 days.


markjjo:
@MSnahuelrey PR updated and then merged. Thanks for submitting this change.

PR #7117, opened 2021-12-07
Updating Diagnostic button and note

salarson:


markjjo:
@salarson PR merged. Thanks for submitting the change and the explanation for using the button. Updated topic will be published by EOD.

PR #7160, opened 2021-12-13
Update allowselfservicepurchase-powershell.md to add new IDs and an explanation

mijeffer:
We're relaunching Project and Visio with trials and new product IDs. We're doing work to make sure that admins who have blocked these products already don't have to do any work by automatically blocking the new IDs for them behind the scenes.


mijeffer:
This needs to go live on 12/15, not earlier

cmcatee-MSFT:
For publishing on 12/15 with 10:30AM build

PR #7188, opened 2021-12-15
Update classifier-how-to-retrain-content-explorer.md

ansaripa:
further to mail thread on deprecation of retraining feature to @chrfox


chrfox:
The edits you made conflict with other notes in the article.  I need to speak to you to understand exactly how retraining works and the scope of accuracy feedback

PR #7199, opened 2021-12-17
Update alert-policies.md

johnengels:
@Dansimp -

Adding the manually triggered user compromise investigation alert, which is going to GA status.


johnengels:
Please update to ensure customers can find info on new alert.

markjjo:
@johnengels PR merged. Thanks for submitting this PR. Also, I changed the severity value to **Medium** because I checked two PROD tenant and the severity rating for this alert policy in both tenants was Medium.  If you change the severity rating in code, please ping me and I will updated the  article or you can submit a PR with the change.

PR #7209, opened 2021-12-17

j0shbregman:
Need to update documentation because we're adding DisableRoutinelyTakingAction to Tamper Protection.


denisebmsft (approved):
Thank you @j0shbregman

PR #7285, opened 2021-12-31
More detail for Do Not Rewrite syntax

JeremyTBradshaw:
Added some bullet points to help people get the syntax right, without needing to do so much trial and error.


JeremyTBradshaw:
I missed escaping a couple of asterisks (*).  Will redo a new PR.

PR #7351, opened 2022-01-10

jkushwah:


yogkumgit:

JohanFreelancer9 (approved):

yogkumgit:
@markjjo Please merge this PR. Thanks

markjjo:
@jkushwah PR merged. I revised your change and moved it to the "conditions guidelines" section as per our IM. 

Any condition that uses an operator with **Contains** and **Equals** logic will return similar search results for simple string searches. A simple string search is a string in the condition that doesn't include a wildcard). For example, a condition that uses **Equals any of** will return the same items as a condition that uses **Contains any of**.

Let me know in Teams if you want me to update or further revise.

PR #7479, opened 2022-01-24
Update TAM to CSAM

adamjbrewer:
Change "Technical Account Manager" to read "Customer Success Account Manager", which is the current title of that role.


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @adamjbrewer

PR #7492, opened 2022-01-25
Endpoint is now supported for entities (both in public preview)

maberzin:
@chrfox  both named entities and endpoint devices are in full public preview


JohanFreelancer9 (approved):

yogkumgit:
@chrfox Please merge this PR. Thanks

PR #7500, opened 2022-01-25
Adding note for respective blocks

dhairyya:
To help users understand how the various blocks block emails in the flow so that they can make the respective block and configure which location the email will land in.


MSFTTracyP (approved):
Thank you!

PR #7536, opened 2022-01-28
Update import-physical-badging-data.md

garajori:
The script link was not pointing to right guidance script, so modified the link and also in a few places "csv" was mentioned instead of "json". Please approve soon as customers are actively trying the connector.


markjjo:
@garajori PR merged. Thanks for updating.

PR #7537, opened 2022-01-28
Update import-physical-badging-data.md

garajori:
Sample Data was not syntaxed correctly


markjjo:
@garajori PR merged. Thanks for updating.

PR #7548, opened 2022-01-29
Connect-ExchangeOnline Modifications

YuriySamorodov:
Removed banner and let Connect-ExchangeOnline import just a single command Get-MailboxFolderStatistics


JohanFreelancer9 (approved):

markjjo:
@YuriySamorodov Thanks for updating. But, what is the reason to make this change? To limit the EXO session to only the one cmdlet?

YuriySamorodov:
Hi @markjjo! Not only that. But also skip banner, which I personally find annoying, especially in scripts.

markjjo:
@YuriySamorodov Thanks for response. PR merged.

PR #7603, opened 2022-02-02
Update supported PS version for AAD Graph Module

juntakata:
As per discussion with Adam Edwards, the Azure AD PowerShell module is only supported on Windows PowerShell 5.1. Users should use Microsoft Graph PowerShell instead which is supported on the very latest versions of PowerShell 7.


yogkumgit:
@JohanFreelancer9 kindly do the copy editor review for this PR. Thanks

JohanFreelancer9 (approved):

PR #7653, opened 2022-02-08
Updating note on status of Built-in protection

dhairyya:
The feature is GA now and the customers are confused because of this note. So, updating it to set the correct expectations here.


MSFTTracyP (approved):
Thank you

PR #7654, opened 2022-02-09
Update office-365-air.md

johnengels:
Removing the separate alert policy table.  We're getting one table updated and the other gets missed on many of the recent changes.  Instead of duplicating a partial alert policy list, we will consolidate it with the default policy list - then add a column related to automated investigations over there.

Also made a slight change to the manually triggered investigations to reflect that these have alerts now.


johnengels:
@Dansimp @MSFTTracyP - some changes we need to make to ensure we get the alert/AIR linkage a bit clearer.  Related change to the alerts documentation coming as well.

PR #7655, opened 2022-02-09
Update alert-policies.md

johnengels:
Updating the alert table to contain the automated investigation information here instead of over in the AIR section.  That way it's one table to update and we avoid doc discrepancies when we make new alerts that have investigations.


johnengels:
@Dansimp @MSFTTracyP - here is the alert documentation change to incorporate the AIR info.

markjjo:
@johnengels PR merged. Thanks for making these changes.

johnengels:
@markjjo - Two questions:
- how long does it typically take to get items updated?
- when I do a pull request, do I need to close the pull request to get the change into the queue?

markjjo:
@johnengels I will reply with answers in email.

PR #7659, opened 2022-02-10
Update create-retention-policies.md

cgill:
When reviewing this as a mentoring opportunity, quickly realized that the Compliance center has been updated to include these options below Information Governance (and the retention policies will only be available under policies after creation). Hoping that you agree these changes help others along the way when configuring (and learning) how to get started with retention policies. Thanks in advance! CG


JohanFreelancer9 (approved):
Copy review - @cgill, looks good but please just confirm the capitalization of the selections "Information **G**overnance" and "Retention **P**olicies" (I do not have access to check). Many thanks.

cabailey:
#sign-off   Yes, agree!

PR #7750, opened 2022-02-18
Setting is deprecated

samkellyd:
Setting is deprecated, maybe we should archive the page?
https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/use-group-policy-microsoft-defender-antivirus?view=o365-worldwide


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @samkellyd

PR #7754, opened 2022-02-18
Update configure-server-endpoints.md

pahuijbr:
Server licensing is different from user licensing and so "P2" does not apply, changed the link text accordingly.


denisebmsft (approved):
Thank you Paul!

PR #7770, opened 2022-02-21
Remove Teams specific port per media type

mibrest-MSFT:
In some cases (VDI, web clients, "add participant" from meeting...) Teams might send audio traffic to another service-side port.

In most cases:
- UDP3478 = Relay discovery allocation and real-time traffic (audio/video/appsharing)
- UDP3479 = Audio
- UDP3480 = Video
- UDP3481 = AppSharing

In the cases highlighted previously, audio might be sent to UDP3480 or UDP3481. As a result, we should no longer mention that each port belongs to a specific media type, as it is not always true, and it generates support cases. We should instead refer to the whole range UDP3478-3481 as a single entity without distinguishing media type per port.


JohanFreelancer9 (approved):

yogkumgit:
@kelleyvice-msft Please merge this PR. Thanks

PR #7780, opened 2022-02-22
Add Update Orchestrator exclude files

yuhonda:
If these update orchestrator files are changed by other process, it occurs unexpected automatic update behavior on server. please consider to add this exclusion.


JohanFreelancer9 (approved):

yuhonda:
I would like to add this exclusions as recommended but I did not confirm that this exclusion is added as default in Defender. Then, I cancel this PR.

PR #7821, opened 2022-02-25
Update create-a-report-on-holds-in-ediscovery-cases.md

edmatosm:
Updated path instructions to make it clearer that user just needs to input target folder. Added code for Advanced eDiscovery cases, since script was only getting Core eDiscovery case holds.


JohanFreelancer9 (approved):

markjjo:
@edmatosm PR merged. Thanks for submitting and for making the requested change to add CaseType.

PR #7834, opened 2022-02-28
Important note on correct use of PowerShell Module

rtmorato:
I have created a change where we need to be clearer on which PowerShell module we should use on this kind of checking. It is normal that users use the SCC powershell to see the logs where in the reality we must use the EXO. This was discussed in one ICM opened.


markjjo:
@rtmorato PR merged. I revised the note and move it. The sentence immediately preceding the cmdlet example says to run the following command in EXO PS. So I moved the note to after the cmdlet example so as not to same the same thing twice in two consecutive sentences.

rtmorato:
Thank you Mark.

PR #7865, opened 2022-03-02
Added Modified by column to Senders, URLs and Files

dhairyya:
Updating documentation after feature change as the customer wants to understand who last modified the entry in TABL for Senders, URLs and files.


JohanFreelancer9 (approved):

MSFTTracyP (commented):
Thanks

PR #8013, opened 2022-03-15

kurtsarens:
Scheduled scans are not supported when Defender is in Passive mode. See
Incident 289114764 : [M365D Escalation] - Microsoft Defender Antivirus (configured in Endpoint Manager) is not triggering scheduled scans on workstations


denisebmsft (approved):
Thank you @kurtsarens

PR #8052, opened 2022-03-17
CA policy update

hathin:
@tiaraquan FYI, added more detail to let customers know that we attempt to exclude ourselves but theres a chance that we may be unsuccessful and what the consequence is.


JohanFreelancer9 (approved):

tiaraquan:
#sign-off

Issue #8009, opened 2022-03-14
PowerShell script enhancement in VDI documentation


brentil:
As written the last line of the PowerShell script could error if $vdmpath is not on the C: drive.  The command assumes it is the C: as part of the cmd command to change to that drive letter.  A better solution is to us the native cd /d command instead.  The /d command changes both drive and directory instead of the basic directory only.  The command I suggest below will work no matter what drive is being used.

As written
`cmd /c "cd $vdmpath & c: & mpam-fe.exe /x"`

Instead use
`cmd /c "cd /d $vdmpath & mpam-fe.exe /x"`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c87fa68a-a881-7ab7-97df-5da73dc56692
* Version Independent ID: 0c031441-cc26-903a-4d53-c044ea790d2d
* Content: [Microsoft Defender Antivirus Virtual Desktop Infrastructure deployment guide](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus?view=o365-worldwide#download-and-unpackage-the-latest-updates)
* Content Source: [microsoft-365/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

msbemba:
@brentil Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/8062

Once the author approves, the changes will be live.

Thanks
Sri

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @brentil Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #8062, opened 2022-03-18
Update deployment-vdi-microsoft-defender-antivirus.md

msbemba:
Made changes to the last line of the PowerShell script could error if $vdmpath is not on the C: drive. The command assumes it is the C: as part of the cmd command to change to that drive letter. A better solution is to use the native cd /d command instead. The /d command changes both drive and directory instead of the basic directory only. The command I suggest below will work no matter what drive is being used.


msbemba:

msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/8009

JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @msbemba @JohanFreelancer9 @yogkumgit

PR #8213, opened 2022-04-05
Merged two lists of blockers for protected documents on the web

bb-froggy:
There were two separate lists of label properties that both result in SPO not being able to process a file and office on the we not being able to open that documents. I merged the two, because the underlying cause is the same.

The only item I have not tested myself is the "User access to content expires".

And user-defined permissions is about to come in June: https://www.microsoft.com/en-us/microsoft-365/roadmap?filters=&searchterms=85619

### DIFF
--- a/microsoft-365/compliance/sensitivity-labels-sharepoint-onedrive-files.md
+++ b/microsoft-365/compliance/sensitivity-labels-sharepoint-onedrive-files.md
@@ -82,7 +82,7 @@ Use the OneDrive sync app version 19.002.0121.0008 or later on Windows, and vers
   - **User access to content expires** is set to a value other than **Never**.
   - **Double Key Encryption** is selected.
   - Encryption that was applied **independently from a label**, for example, by directly applying a Rights Management protection template.
-  - Encryption that uses an on-premises key ("**hold your own key**" or HYOK)
+  - Encryption that uses an on-premises key ("**hold your own key**" or HYOK).
 
     For labels with any of these encryption configurations, the labels aren't displayed to users in Office for the web. Additionally, the new capabilities can't be used with labeled documents that already have these encryption settings. For example, these documents won't be returned in search results, even if they are updated. Documents that have been encrypted in one of these ways can't be opened in Office for the web.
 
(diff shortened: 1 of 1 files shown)


microsoft-365/compliance/sensitivity-labels-sharepoint-onedrive-files.md
```diff
@@ -81,8 +81,10 @@ Use the OneDrive sync app version 19.002.0121.0008 or later on Windows, and vers
   - **Let users assign permissions when they apply the label** and the checkbox **In Word, PowerPoint, and Excel, prompt users to specify permissions** is selected. This setting is sometimes referred to as "user-defined permissions".
   - **User access to content expires** is set to a value other than **Never**.
   - **Double Key Encryption** is selected.
+  - Encryption that was applied **independently from a label**, for example, by directly applying a Rights Management protection template.
+  - Encryption that uses an on-premises key ("**hold your own key**" or HYOK)
```
JohanFreelancer9:
```suggestion
  - Encryption that uses an on-premises key ("**hold your own key**" or HYOK).
```

JohanFreelancer9 (changes requested):
Copy review - @bb-froggy, kindly commit the minor suggested change. Many thanks.

JohanFreelancer9 (approved):

cabailey:
@bb-froggy I appreciate why you thought they could be combined but these two lists were kept separate because they fall into different categories, even though both result in being unable to open in Office on the web: 

The first is for encryption applied with sensitivity labels, which is why they have the configuration link before the list so that customers can change if required. The link wouldn't work for the second list (DKE can also be applied via the SDK). They also have the explanation that users don't see these labels to apply (the most frequently reported impact from customers). When it says "the new capabilities can't be used with labeled documented that already have these encryption settings", this includes opening them in Office on the web, populating the Sensitivity column, and search. Search is the one that customers seem to forget for this one, which is why it's specifically called out. Since I work with the engineering team, I'll be sure to update the list when any of these limitations are addressed with new releases, but we don't do that until they become publicly available. 

The other list is about encrypting independently from sensitivity labels, which is less common & much less frequently reported by customers (why later in the list), where the main symptom reported is that they can't be opened in Office on the web. With such a long list of limitations, we work hard to order the most frequently reported symptoms higher in the list.

bb-froggy:
Okay, I understand the rationale now.

My concern is that a reader might see only one of the two lists, not expecting that there is a second one, and assuming that the one list is comprehensive. IMHO, the wording doesn't make it 100 % clear that the first list is for encryption applied through a Sensitivity Label, while the other list refers to encryption applied through other means.

Maybe rephrase the wording for the second list to something like "Documents that have been encrypted without corresponding Sensitivity Label can't be opened in Office for the web, e.g." and then the actual list?

PR #8226, opened 2022-04-06
Added note that CP will be on when TP is on

j0shbregman:
The documentation needs to be updated to clarify the fact that when TP is turned on via MDE portal, that CP must be on.  This is done to ensure that TP can be disabled via CP - otherwise there can be a race condition.  This change was made at the end of last year in MoCAMP 2111, so need to update documentation to reflect this change.


j0shbregman:
@denisebmsft This is good to go.

denisebmsft (approved):
Thank you @j0shbregman

PR #8246, opened 2022-04-07
Update compare-mdb-m365-plans.md

james-maclean:
Add the fact that its purely referring to standalone in the second table as the document is confusing.


denisebmsft (approved):
Thank you @Dope-Lemon

PR #8320, opened 2022-04-14
New Case action - Manage pseudonymize

ItaloDuarte:
There is a new case action called Manage pseudonymize and there is no information about this action.
we would like to have information about it on this page, because it is already explaining all actions here.
Thank you


yogkumgit:
@robmazz Please merge this PR. Thanks

robmazz:
@ItaloDuarte - thanks for the feedback. I'm unaware of this case action, but will confirm with the PG team. If a section is needed, we will add it. Thanks again.

Issue #8318, opened 2022-04-14
Broken link


robelliottuk:
[Enter feedback here]

This part of the doc has a broken link:
"For more information, follow the guidance in [Deployment guide for Microsoft Defender Antivirus in a virtual desktop infrastructure (VDI) environment](https://docs.microsoft.com/en-us/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus)."

This link isn't available:
https://docs.microsoft.com/en-us/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ef5b23fe-37ce-e3b0-ad76-55a416eaa865
* Version Independent ID: 26becb70-47ff-38fe-592f-99d84f6aba12
* Content: [Onboard non-persistent virtual desktop infrastructure (VDI) devices](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/configure-endpoints-vdi?view=o365-worldwide)
* Content Source: [microsoft-365/security/defender-endpoint/configure-endpoints-vdi.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/security/defender-endpoint/configure-endpoints-vdi.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @mjcaparas
* Microsoft Alias: **macapara**

RAJU2529:
@yogkumgit , full link is working 
**https://docs.microsoft.com/microsoft-365/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus?view=o365-worldwide**
I replaced above link in article

PR #8325, opened 2022-04-15
broken link fixed

RAJU2529:
as per user #8318, so i fixed broken link with correct link

only the full link is working 
**https://docs.microsoft.com/microsoft-365/security/defender-endpoint/deployment-vdi-microsoft-defender-antivirus?view=o365-worldwide**


denisebmsft (approved):
Thank you @RAJU2529

PR #8348, opened 2022-04-19

adysss22:
At line 192, I have removed the entry "| Command Prompt |  1. On a Windows device, open Command Prompt.<br/>2. Type `sc query windefend`, and then press Enter.<br/>3. Review the results to confirm that Microsoft Defender Antivirus is running in passive mode.  | " because when using the mentioned method to query the windefend service you cannot see if Microsoft Defender Antivirus is running in passive or active mode. You can only see if the service is running or stopped when running the sc query windefend command.

s1
.


denisebmsft (approved):
Thank you @adysss22

PR #8355, opened 2022-04-19
Multiple email attachments not supported

dhairyya:
Getting escalations as email submissions from third party add-ins are rejected due to multiple email attachments as we only support one email attachment


MSFTTracyP (approved):
Thanks

PR #8360, opened 2022-04-19
Section not relevant

dhairyya:
Removed the section Turn off the built-in reporting experience as the ReportJunkEmailEnabled setting no longer needs to be disabled as per discussion with Engineering.


MSFTTracyP (approved):
PM Verified. Approving.

PR #8382, opened 2022-04-21
Updating roles and groups for spoof intelligence

dhairyya:
Two documents have different requirements from TABL document which is confusing customer so bringing those documents to the same


MSFTTracyP (approved):
Thanks

PR #8417, opened 2022-04-26
Update office-365-worldwide-endpoints.md

wenfMS:
Hi team,

We are IC3 Data Platform team. Our service host is end with "*.teams.microsoft.com" and we need to update 2 IP address to allow customer visit our service.

Thanks!


wenfMS:
Our IP address might change and will update this doc later.

PR #8421, opened 2022-04-26
Changed Cosoms to Cosmos

sguitardude:


cabailey:
Thanks for catching this typo!

PR #8424, opened 2022-04-27
Location update for SE

hathin:
SE is now active in Romania. Updated privacy doc to reflect this. @tiaraquan FYI


tiaraquan:
#sign-off

Issue #8262, opened 2022-04-08
Consistent terminology


ma261065:
The section "How long it takes for retention policies to take effect" has a diagram that uses the term "synced" The text underneath talks about "distributed". Are they the same thing?

Also, the actual portal shows the status as "Enabled (Success)", while the diagram in the doco shows a status of "Success (ON)". Are they the same thing?

Document Details
⚠ Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.

ID: fd61abe1-251f-1409-d386-145d93c8f18f
Version Independent ID: b340c591-d6bf-ae5f-e9b1-b32be18d6bee
Content: [Verify eligibility for Microsoft 365 Education subscriptions](https://docs.microsoft.com/en-us/microsoft-365/commerce/subscriptions/verify-academic-eligibility?view=o365-worldwide#verify-your-schools-academic-eligibility)
Content Source: [microsoft-365/commerce/subscriptions/verify-academic-eligibility.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/commerce/subscriptions/verify-academic-eligibility.md)
Service: o365-administration
GitHub Login: @cmcatee-MSFT
Microsoft Alias: cmcatee

msbemba:
@ma261065 Thank you for your feedback. I do not see the section "How long it takes for retention policies to take effect" in this document. Could you please provide the URL of the document you are referring to ?

Thanks
Sri

ma261065:
Sorry, this is the document - [https://docs.microsoft.com/en-us/microsoft-365/compliance/create-retention-policies?view=o365-worldwide#how-long-it-takes-for-retention-policies-to-take-effect](url)

yogkumgit:
@msbemba

msbemba:
@ma261065 Thank you for your feedback. I will fix the document shortly.

msbemba:
@ma261065 Thank you for your feedback. We have fixed the diagram to match the portal. This can be tracked via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/8429

Once the author approves, the changes will be live.

Thanks
Sri

yogkumgit:
Closing us per Author's input. Thanks

PR #8429, opened 2022-04-27
Update retention-policy-timings.png

msbemba:
Changed diagram to fix terminology (Consistent)

synced is changed to distributed and the status is changed to Enabled (Success) as seen in the portal


msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/8262

msbemba:

JohanFreelancer9 (approved):

cabailey:
Investigating

cabailey:
These are good suggestions now the UI is updated, but I couldn't accept the file because it doesn't meet our accessibility requirements and would be inconsistent with the other diagrams for retention label timings. These are now updated, so closing this PR. For future reference, this would have been better as a GitHub Issue.

Issue #8433, opened 2022-04-27
'Microsoft 365 licensing guidance for security & compliance' removed


rramoscabral:
At retention-regulatory-requirements.md (`https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/compliance/retention-regulatory-requirements.md`) the link  >*[Microsoft 365 licensing guidance for security & compliance](/office365/servicedescriptions/microsoft-365-service-descriptions/microsoft-365-tenantlevel-services-licensing-guidance/microsoft-365-security-compliance-licensing-guidance)` is not working. I can not find the markdown file. 

The same error will get at `https://docs.microsoft.com/en-us/office365/servicedescriptions/microsoft-365-service-descriptions/microsoft-365-tenantlevel-services-licensing-guidance/microsoft-365-security-compliance-licensing-guidance`

rramoscabral:
The links is working. The Maybe it was a temporary issue, as per the feedback from Carol Bailey.

PR #8437, opened 2022-04-27
link changed

RAJU2529:
as per user report #8433, so i added full link
**https://docs.microsoft.com/en-us/office365/servicedescriptions/microsoft-365-service-descriptions/microsoft-365-tenantlevel-services-licensing-guidance/microsoft-365-security-compliance-licensing-guidance**


cabailey:
I can't accept this PR - we're not supposed to use full URL links to other pages on docs.microsoft.com.  Are you sure the existing link not working wasn't a temporary issue?  It works for me.

JohanFreelancer9 (approved):

JohanFreelancer9 (changes requested):
@RAJU2529, please see @cabailey's comment. Thanks.

RAJU2529:
@cabailey . I checked with half link, it does not worked, so i changed to full link, then it's worked. So please  check from your side.

cabailey:
@RAJU2529 It's not a "half link" but a relative (site) link within our doc set.  There's no need to add "https://docs.microsoft.com" because the source page is also published on docs.microsoft.com.  We're asked to use relative links like this because they are faster to load (the browser doesn't need to resolve the domain name again but instead redirect to a page from the same source) and relative links continue to work in air-gap scenarios.  You can read more about this linking requirement here: https://docs.microsoft.com/en-us/contribute/how-to-write-links#links-from-one-article-to-another  

I've checked this multiple times and this link on this published page always works for me.  When I hover over the link, I can see the full URL displayed in the browser status bar, and it's correct.  We use exactly the same link on multiple pages and not heard of this being a problem before.

I'm not sure what to suggest at this point if it's not working for you, but a PR with the full URL to docs.microsoft.com will be blocked from merging because it's against our publishing rules.

RAJU2529:
@cabailey . Please give me link of published page,  i checked links in GitHub page only so it's not working.

cabailey:
@RAJU2529   That would explain it.  None of these documentation pages are designed to be consumed directly in GitHub but on the Microsoft docs.microsoft.com website (or you'll have this problem with many links).  For this specific page: https://docs.microsoft.com/microsoft-365/compliance/retention-regulatory-requirements 

Closing this PR.

RAJU2529:
@cabailey, in that page, link is working , Thanks for the link.Sorry for the disturbance form my side

PR #8464, opened 2022-05-02
Fixed issues with sample regexes

E-Saggese:
The regex given in the example didn't match the example values given, so it would not detect them as stated. 
Also, using a dictionary instead of a regex is becoming a more popular alternative than a regex for multi-word evidence scenarios, so updated the explanation to center on that alternative.


microsoft-365/compliance/sit-get-started-exact-data-match-test.md
```diff
@@ -79,8 +79,8 @@ If you don't find any matches, here are some troubleshooting tips.
...
+|An EDM SIT with primary elements and no secondary elements defined detects items, but doesn't detect, or detects fewer matches than expected, when primary and secondary elements are required.  | If values in a column used for secondary evidence are not composed of a single word or strings that don't contain spaces, commas or other word separators, you will need to associate them with a sensitive info type that uses either a REGEX designed to to detect multi-word strings that follow the desired pattern (e.g. a fixed number of consecutive words that start with an uppercase character), or a keyword dictionary that lists all the unique values in that column. For example, if there's an additional evidence column for a person's city or residence, you can create a list with all the unique city names from the table and use it to create a dictionary-based sensitive information type. Use this SIT as the classification element for the corresponding column in your EDM sensitive info type by exporting and editing the EDM SIT definition in XML. See [Create a rule package manually](sit-get-started-exact-data-match-create-rule-package.md#create-a-rule-package-manually)|
```
JohanFreelancer9:
```suggestion
|An EDM SIT with primary elements and no secondary elements defined detects items, but doesn't detect, or detects fewer matches than expected when primary and secondary elements are required.  | If values in a column used for secondary evidence are not composed of a single word or strings that don't contain spaces, commas, or other word separators, you will need to associate them with a sensitive info type that uses either a REGEX designed to detect multi-word strings that follow the desired pattern (e.g. a fixed number of consecutive words that start with an uppercase character), or a keyword dictionary that lists all the unique values in that column. For example, if there's an additional evidence column for a person's city or residence, you can create a list with all the unique city names from the table and use it to create a dictionary-based sensitive information type. Use this SIT as the classification element for the corresponding column in your EDM sensitive info type by exporting and editing the EDM SIT definition in XML. See [Create a rule package manually](sit-get-started-exact-data-match-create-rule-package.md#create-a-rule-package-manually).|
```

JohanFreelancer9 (changes requested):
Copy review - @E-Saggese, kindly commit the additional suggested change to correct a few typos. Many thanks.

JohanFreelancer9:
@E-Saggese, thank you for committing the suggested changes, but I see that you have closed the PR. Was that intentional?

E-Saggese:
Apologies, closing the request was unintentional.

JohanFreelancer9 (approved):

PR #8517, opened 2022-05-06
Update user-tags.md

faoquong:
@chrisda can we make it call into this URL https://docs.microsoft.com/en-us/microsoft-365/security/office-365-security/view-email-security-reports?view=o365-worldwide#top-senders-and-recipients-report


chrisda:
@faoquong, done.

Issue #8666, opened 2022-05-19
Bad path in example


pensivepaddle:
Wrong path in "Issue: Tamper protection is reported as disabled" section

`$ sudo grep -F '\[{tamperProtection}\]: Feature state:' /Library/Logs/Microsoftmdatpmicrosoft_defender_core.log \| tail -n 1`

should probably be 
`$ sudo grep -F '\[{tamperProtection}\]: Feature state:' /Library/Logs/Microsoft/mdatp/microsoft_defender_core.log | tail -n 1`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b466be31-a444-e4f0-f56e-1b514776df63
* Version Independent ID: b466be31-a444-e4f0-f56e-1b514776df63
* Content: [Protect macOS security settings with tamper protection](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/tamperprotection-macos?view=o365-worldwide)
* Content Source: [microsoft-365/security/defender-endpoint/tamperprotection-macos.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/security/defender-endpoint/tamperprotection-macos.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @mjcaparas
* Microsoft Alias: **macapara**

RAJU2529:
@mjcaparas Please verify PR #8668  and merge if correct.

yogkumgit:
@RAJU2529 Thank you very much for the contribution and updating the documentation with PR. @pensivepaddle Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #8668, opened 2022-05-19
added backslash

RAJU2529:
as per user report #8666 , so i added backslash

Screenshot 2022-05-19 181959


denisebmsft (approved):
Thank you @RAJU2529

PR #8677, opened 2022-05-19
Updated named entities / DLP supported workloads to include Microsoft Defender for Cloud Apps

maberzin:
@chrfox - Named entities are now fully supported in the DLP solution for the Microsoft Defender for Cloud Apps workload (commercial cloud only; named entities are not yet available in the 3 gov clouds).


yogkumgit:
@JohanFreelancer9

JohanFreelancer9 (approved):

PR #8693, opened 2022-05-23
Remove single user in browser session requirement

oschreib:
Microsoft 365 Defender now supports accessing multiple tenants in the same browser session.


Dansimp:
moved this change to the private repo. Closing this PR

PR #8694, opened 2022-05-23
Update Access to Enable Archive

Charli-Rose:


cabailey:
No need to use the Actions menu when the Enable archive option is at the top of the page, as shown in the screenshot for step 3.  Do you not see this option your tenant?  From what I see in the tenants that I test with, accepting this PR would make the screenshot inaccurate & also the note about using search out of order.  If you prefer to use the Actions menu, that's fine but not our first choice for documented instructions.

Charli-Rose:
Understand. No, my current test admin does not have the ability to do this from the horizontal menu. I'll check again to be sure.

cabailey:
@Charli-Rose  Yes, please do check - sounds like a bug to me if you don't see it after selecting a user as per the instructions & screenshot. You should see either "Enable archive" (I fixed the casing to match the UI after checking this myself, which is why this PR now has a merge conflict) or "Disable archive" appear after the Refresh option. Note that you don't see this option until a user is selected.

Charli-Rose:
@cabailey The option that is described in the original document is now available.

cabailey:
@Charli-Rose  Somebody pointed out to me today that this option doesn't have an icon, which is unusual.  So it's possible not displaying might be down to a resolution issue, or a small window.  Similar options lose the display name when the window shrinks, but leaves the icon.  With no icon, we were wondering if it might "disappear" and this explains why you couldn't see it..

We're going to look into this but if you can reproduce it, please do report it to Support so this can go through official channels.  It might be a simple fix by adding an icon.

PR #8742, opened 2022-05-27
Update microsoft-365-mailbox-utilization-service-alerts.md

holliep1:
I've changed all reference of 'alert' to advisory (or advisories) and updated the duration from four days to seven days.  I originally created this article with Mark Johnson.


holliep1:
None

PR #8795, opened 2022-06-02
Update broken link in TOC.yml

gatesry:
Removing the broken link to the compliance section. On the Wayback machine, I found the old article but couldn't locate it within the MS Compliance repo.  @robmazz - Including you for visibility. 

image

image


robmazz:
@gatesry - thanks for the feedback, closing this and will fix (along with some other changes) on a separate PR.

PR #8891, opened 2022-06-13
Update import-hr-data.md

NarraSaiKrishna:
Updating the document based on the feedback received from the team


v-tophillips (approved):

NarraSaiKrishna:
Hey, I think some wrong commit happened because I don't see any changes in the public document.

v-tophillips:
I've verified this (I don't see it in live docs, either) and will fix this in a separate PR. Please check again after 24 hours.

v-tophillips:
After investigating further, it looks like the changes were made but may not have appeared when you checked. The changes are, in fact, in the published files and should have now replicated to the published docs.

PR #8954, opened 2022-06-20
Update user-submission.md

cammurray:
Add reference to condition customer can get in to if they are performing phishing simulations & using a submissions mailbox.


cammurray:
@dhairyya fyi

Issue #8926, opened 2022-06-15
"continuously collects behavioral cyber telemetry" -- but not always


cronq:
After working with Microsoft support when we saw a large number of gaps in event data in MDE, an MDE engineer said this is working as designed. From the support ticket "MDE is not intended to be an auditing/logging solution that records every operation or activity that happens on a given machine...Our sensor has an internal throttling/capping mechanism, so the high rate of repeat identical events won't flood the logs...when we transmit telemetry to the cloud". This, or something similar to it, should be noted on this page (and likely others as well). Any reader will make the assumption that ALL events will be sent to MDE with the existing statement "This includes process information, network activities, deep optics into the kernel and memory manager, user login activities, registry and file system changes, and others. The information is stored for six months, enabling an analyst to travel back in time to the start of an attack." Thus, a note about the throttling, likely event loss, and MDE is "not intended to...record every operation" is VERY relevant to users and potential users. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 072bd7c6-9f5b-4de2-2d65-f83e7adbc85f
* Version Independent ID: 39c90a0c-6bce-68a3-7b1f-138819744b76
* Content: [Overview of endpoint detection and response capabilities](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/overview-endpoint-detection-response?view=o365-worldwide)
* Content Source: [microsoft-365/security/defender-endpoint/overview-endpoint-detection-response.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/security/defender-endpoint/overview-endpoint-detection-response.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @mjcaparas
* Microsoft Alias: **macapara**

yogkumgit:
@msbemba

msbemba:
@cronq Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/9030

Once the author approves, the changes will be live.

Thanks
Sri

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @cronq Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #9030, opened 2022-06-27
Update overview-endpoint-detection-response.md

msbemba:
Updated note as per the microsoft support ticket. Defender for Endpoint detection is not intended to be an auditing or logging solution that records every operation or activity that happens on a given machine.Our sensor has an internal throttling mechanism, so the high rate of repeat identical events will not flood the logs.


msbemba:

msbemba:
Issue#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/8926

denisebmsft (approved):
Thank you @msbemba

JohanFreelancer9 (approved):

PR #9032, opened 2022-06-27
Update enable-troubleshooting-mode.md

kurtsarens:
Simplifying queries. We had received complaints from customers that the queries do not work
Incident 312104959 : [MDATP/MTP - CSS Escalation ] - AH query for Troubleshooting Mode not working


denisebmsft (approved):
Thank you @kurtsarens

PR #9067, opened 2022-07-01
Dynamic delivery needs to be off

dhairyya:
Dynamic delivery for custom mailbox needs to be off otherwise if email is submitted when the email attachment is being scanned, we provide an error as the email attachment is being encrypted.


dhairyya:

JohanFreelancer9 (approved):

PR #9180, opened 2022-07-14
Added edge case for TABL sync

dhairyya:
Some tABL entries can take upto 24 hours instead of 30 mins due to which we are getting escalations


dhairyya:
@chrisda @MSFTTracyP could either of you please approve this for me?

PR #9182, opened 2022-07-15
Added edge case for TABL sync

dhairyya:
Some TABL entries can take up to 24 hours instead of 30 mins due to which we are getting escalations


dhairyya:
@chrisda or @MSFTTracyP could either of you please approve this for me?

PR #9204, opened 2022-07-20
Update create-sensitivity-labels.md

rtmorato:
I have updated the lines 125 and 126.


cabailey:
For labels that you apply to Teams, this limitation is documented here: https://docs.microsoft.com/en-us/microsoft-365/compliance/sensitivity-labels-teams-groups-sites

Sensitivity labels have a lot of different settings and it's highly advisable to check they work as expected for the apps and services your users use before deploying in production.  Check the documentation specific to the scenario as well.  We can't track and list all the exceptions here, and to list only some could imply support for all the others.

PR #9253, opened 2022-07-26
Updated references from original creation

loryanstrant:
A few things have changed from when I originally authored this article two years ago.
Several references have been updated:
- OneNote is no longer "OneNote 2016"
- There is a differentiation between classic and modern workspaces in Power BI and their association with M365 Groups
- Power Automate can be used to migrate SharePoint list contents


MikePlumleyMSFT:
Thanks, @loryanstrant!

PR #9321, opened 2022-08-03
Safe domains are honored

dhairyya:
Checked with engineering. There is a typo as the behavior has never been changed.


dhairyya:

PR #9397, opened 2022-08-11
Updated sender to domains and email addresses

dhairyya:
Added callout for recipient which was added to the sender tab in TABL causing its name to be updated to domains and email addresses


dhairyya:
@chrisda and @MSFTTracyP could you please approve this for me?

denisebmsft (approved):
Thank you @dhairyya

PR #9418, opened 2022-08-12
Update dlp-on-premises-scanner-get-started.md

aamalhot:
should be on-premises instead of Endpoint DLP since this is on-prem DLP doc


aamalhot:
@chrfox please review

JohanFreelancer9 (approved):

PR #9532, opened 2022-08-26
Update wrong reference on automation-levels.md

BenjaminLam:
Updating such that Defender for Business isn't compared to Defender for Business, but to Defender for Endpoint insted


JohanFreelancer9 (approved):

denisebmsft (approved):
@BenjaminLam great catch! THANK YOU!

PR #9572, opened 2022-08-30
Updating Doc's per product groups recommendation

edberei:
When disabled or uninstalled, Microsoft Defender Antivirus isn't used as the antivirus app. Files aren't scanned and threats aren't remediated. Disabling or uninstalling Microsoft Defender Antivirus isn't recommended in general; if you're using a non-Microsoft antimalware/antivirus solution and the device is onboarded to Microsoft Defender for Endpoint, Microsoft Defender Antivirus **must** be in passive mode.


edberei:
@opbld32 Closing the request until we have PM sign off.

PR #9635, opened 2022-09-07
Updated document error

dhairyya:
Updated block and block not matched scenario for example of contoso.com


dhairyya:
@MSFTTracyP and @chrisda could you please approve it?

PR #9661, opened 2022-09-10
Called out top level domain with *

dhairyya:
Added call out for topleveldomain/* to mitigate customer escalations as seen for UVS


dhairyya:
@chrisda and @MSFTTracyP could you please approve this for me?

JohanFreelancer9 (approved):

PR #9773, opened 2022-09-20
Removed URL and file submission note

dhairyya:
As per User Story 2802583, Fedramp customers will be able to see URL and email attachment tab. There data will still not leave the tenant. So removing the notes to avoid confusion.


dhairyya:
@chrisda or @MSFTTracyP could you please approve this change?

JohanFreelancer9 (approved):

PR #9791, opened 2022-09-21
Clarified that basic ASR rules

j0shbregman:
Basic ASR rules should be in block mode.  That's the recommendation.

@denisebmsft


j0shbregman:
@denisebmsft I think this is ready to go

denisebmsft (approved):
Thank you @j0shbregman

PR #9812, opened 2022-09-23
Update device-onboarding-overview.md

ancmarti:
Removing the extra parenthesis from line 79. It doesn't look good in the final format


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @ancmarti. Good catch!

PR #9823, opened 2022-09-26
Update gov.md Device Health Reports GA

JesseEsquivel:
Device health reports are now GA in GCC, GCC-H, and DoD!


denisebmsft (approved):
Thank you @JesseEsquivel

PR #9825, opened 2022-09-27
Update configure-server-endpoints.md

JonathanBrockhausen:
The flag -noETW doesn't exist in the install.ps1 script anymore. It appears this was changed to -NoMSILog. The documentation here should be changed to reflect this as the script will otherwise produce an error. See https://github.com/microsoft/mdefordownlevelserver/blob/main/Install.ps1#L55


JohanFreelancer9 (approved):

yogkumgit:
@mjcaparas Please merge this PR. Thanks

PR #9880, opened 2022-10-04
Added exceptions in DLP for user reports

dhairyya:
Getting escalations that user reports are not visible, due to them getting quarantined by DLP.


dhairyya:
@chrisda could you please review this and approve this?

PR #9930, opened 2022-10-11
Update about-shared-mailboxes.md

vishnutj:
There is not called as "automatic retention policies" instead it should mention "Retention policies" from Compliance portal.


vishnutj:
@chrisda

PR #9981, opened 2022-10-18
Update mac-install-with-intune.md

Dekor86:
The document I believe is outdated, refers to using XML files and deploying with Device Configuration profiles rather than the new Endpoint Security profiles.  Also, multiple settings now available in settings catalog profile types vs existing "Templates" based Device Configuration profiles. Overview table contains links to other articles that are then not mentioned in the steps below, causing confusion on best practice for deploying Defender to macOS.


JohanFreelancer9:
@Dekor86, many thanks for your submission. However, please use the "Send feedback about this page" button instead of "Edit this document" to provide feedback/comments about the article.
"Edit this document" is to propose a direct update to the article. Thanks again.

PR #10017, opened 2022-10-20
Switching references to insider slow to External

goodboyrobot:
Insider slow is not actually a valid name, external is.


bweston-8 (approved):
looks good. Please publish

PR #10077, opened 2022-10-27
Update support-contact-info.md

chenxizhang:
This is Ares from M365 China PM team, we need to make a minor change of the support hours information on the support page


chenxizhang:
@cmcatee-MSFT please help review and approve. thanks

PR #10115, opened 2022-11-01
Update difference-between-document-understanding-and-form-processing-…

seansquires:
…model.md

add training method column to connect ux w/ model type names


tomresing:
I'm going to close this one and open a new one so I can get an additional reviewer on

PR #10138, opened 2022-11-03
removed OS version dependency requirements

billmcilhargey:
Block at first sight no longer has OS dependency requirements other than it works on just Windows OS and running latest antimalware platform for workstation or server


denisebmsft (approved):
Thank you @computeronix

PR #10141, opened 2022-11-03
Update mac-What'snew.md

mmkholy:
Added a note on a known issue with an Apple release :
 " Known issues

Apple has identified an issue on macOS Ventura upgrade, and expected to be fixed in the next release. The issue impacts Microsoft for endpoint security extensions, and might result in losing Full Disk Access Authorization, impacting its ability to function properly.

To mitigate the risk with this issue, ensure that Microsoft for Endpoint has Full Disk Access Authorization."


denisebmsft (approved):
Thank you @mmkholy

PR #10274, opened 2022-11-15

JuanGasca:
Updated document based on MC411428 - After 30 days any inactive mailbox which are in soft-deleted state are permanently deleted and are no longer recoverable.


JohanFreelancer9 (approved):

cabailey:
@JuanGasca Yes, it does seem this was changed recently and not yet updated in the documentation.  Thank you for offering to update the docs - however, there's also a reference to the 183 days in the linked-to More info. So we don't introduce an inconsistency, please update this section as well.

cabailey:
Fixed with https://github.com/MicrosoftDocs/microsoft-365-docs/pull/10295 - closing

PR #10295, opened 2022-11-17

JuanGasca:
Updated document based on MC411428 - After 30 days any inactive mailbox which are in soft-deleted state are permanently deleted and are no longer recoverable.


JohanFreelancer9 (approved):

cabailey:
#sign-off  Merging

PR #10299, opened 2022-11-17
Update attack-surface-reduction-rules-reference.md

baardhermansen:
Corrected PowerShell command for setting a rule in Warn mode.  As it was written, it would produce an error if run, both in PowerShell and in a cmd window.

If invoked from a cmd window, a working command would be
`powershell -command "Add-MpPreference -AttackSurfaceReductionRules_Ids 56a863a9-875e-4185-98a7-b882c64b5ce5 -AttackSurfaceReductionRules_Actions Warn"`


bweston-8 (approved):
This change looks correct to me.

PR #10300, opened 2022-11-17
Update attack-surface-reduction-rules-deployment-implement.md

baardhermansen:
Very small correction; removed a quotation mark.
Capitalized two markdowns for Tip and Note.


bweston-8 (approved):
Good revisions. Thanks!

PR #10343, opened 2022-11-22
Fixed the update date from 2020 to 2023

tduarte14:
Date was wrong here: "(...) and your billing statements will update on or after April 21, 2020", should be "(...) and your billing statements will update on or after April 21, 2023"


tduarte14:
Noticed now that the news is actually from 2020
https://www.microsoft.com/en-us/microsoft-365/blog/2020/03/30/new-microsoft-365-offerings-small-and-medium-sized-businesses/

PR #10518, opened 2022-12-15
Removes SFBO/Hybrid mentions - cleaned targets

mibrest-MSFT:
Cleared up records names and purposes
Removed mentions of SFBO and hybrid from the text
Points to the correct documentation for DNS entries in hybrid environments Cleaned the target column to be more explicit (as it was before the PR of 20th October 2022)


denisebmsft (approved):
Thank you @mibrest-MSFT

PR #10565, opened 2022-12-24
Update multi-tenant-people-search.md

ArvindHarinder1:
The feature name is incorrect and that link is not publicly accessible. It should not be in a public facing document.


ArvindHarinder1:
#sign-off

JohanFreelancer9 (approved):

Issue #10583, opened 2022-12-30
Adding detail to the effects of adding a folder to "protected folders" in ransomware protections settings


Reynolds-Beta:
The information imparted in the page doesn't explicitly indicate whether or not adding a folder also includes sub-folders or not. This small detail has important downstream consequences depending upon the answer.

Please add whether or not sub-folders are included in protection when you add a new folder to the list.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 14fb0d02-ceff-61d8-3041-13df9513769a
* Version Independent ID: 7c98a316-de58-520a-a6ad-06ddaf0d71b3
* Content: [Protect important folders from ransomware from encrypting your files with controlled folder access](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/controlled-folders?view=o365-worldwide)
* Content Source: [microsoft-365/security/defender-endpoint/controlled-folders.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/security/defender-endpoint/controlled-folders.md)
* Service: **microsoft-365-security**
* Sub-service: **mde**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

msbemba:
@Reynolds-Beta Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/10611

Once the author approves, the changes will be live.

Thanks
Sri

yogkumgit:
@msbemba Thank you very much for the contribution and updating the documentation with PR. @Reynolds-Beta Hope this update is helpful for you. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #10611, opened 2023-01-06
Update controlled-folders.md

msbemba:
Added a note that a sub folder is also included in protection when you add a new folder to the list.

fixes#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/10583


msbemba:

PR #10644, opened 2023-01-10
Update switch-to-mde-troubleshooting.md

bracherp:
Adding the correct path for MDE.


denisebmsft (approved):
Thank you so much @bracherp

PR #10723, opened 2023-01-17
Allows cannot be extended

dhairyya:
Removed the part that allows can be extended.


dhairyya:
@chrisda could you please approve these changes?

PR #10748, opened 2023-01-19
Sharing policy needs to be enabled

tudorh:
Enabled property to the sharing policy output as it is required to be enabled - Bookings with me doesn't work without it! Also corrected a type where the last "N" was missing from Anonymous:SharingPolicyAction


yogkumgit:
@JohanFreelancer9

Issue #10692, opened 2023-01-13
Add additional licensing requirements information


jhic:
Hello,

Any chance you can add a fuller description of the licensing requirements for the licensing requirements under this anchor point? 

https://learn.microsoft.com/en-us/microsoft-365/solutions/manage-devices-with-intune-overview?view=o365-worldwide#implementing-the-layers-of-protection-on-and-for-devices

Business Premium should satisfy some, but not all, of the conditions for the 7 steps. Would be handy to have Business plans considered for licensing.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 8ea54678-1318-1a45-987f-91261714fdbb
* Version Independent ID: 8ea54678-1318-1a45-987f-91261714fdbb
* Content: [Manage devices with Intune](https://learn.microsoft.com/en-us/microsoft-365/solutions/manage-devices-with-intune-overview?view=o365-worldwide)
* Content Source: [microsoft-365/solutions/manage-devices-with-intune-overview.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/solutions/manage-devices-with-intune-overview.md)
* Service: **o365-solutions**
* GitHub Login: @BrendaCarter
* Microsoft Alias: **bcarter**

msbemba:
@jhic Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/10758

Once the author approves, the changes will be live.

Thanks
Sri

msbemba:
@jhic Greetings, Our Business Premium writer took a look at this addition and we agree that it could be confusing to add the licensing to this table, especially since the steps are different for connecting Intune to Defender for Business (instead of Defender for Endpoint). Instead, we're taking the feedback and planning to write a separate solution specifically for Business Premium customers and also including Lighthouse partners.

@yogkumgit This can be closed.

yogkumgit:
@msbemba Thank you very much for the contribution and sharing this explanation. @jhic Hope this comment is helpful for you. If you see a documentation update is required, please feel free to open an issue for the same. We proceed here to close it. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

PR #10758, opened 2023-01-20
Update manage-devices-with-intune-overview.md

msbemba:
Added Business premium as some steps are possible with M365 Business premium

fixes#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/10692


msbemba:

BrendaCarter:
Checking with our Business Premium writer before rolling this out . . .

JohanFreelancer9 (approved):

BrendaCarter:
Greetings, Our Business Premium writer took a look at this addition and we agree that it could be confusing to add the licensing to this table, especially since the steps are different for connecting Intune to Defender for Business (instead of Defender for Endpoint). Instead, we're taking the feedback and planning to write a separate solution specifically for Business Premium customers and also including Lighthouse partners.

Issue #10836, opened 2023-01-30
CIS Product Info and Linking


mialavada:
The hyperlinked text “CIS Implementation Group 1, Group 2, Group 3” should be changed to “CIS Critical Security Controls Implementation Group (IG) 1, IG 2, IG 3”, if possible. The hyperlink on that text should also be removed, please. The second bullet should be changed to "CIS Microsoft 365 Foundations Benchmark Level 1 and 2" and should link to the URL that is currently being used for the first bullet. Please and thank you!

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: a406f24f-e0c1-9143-9130-ce0459a66235
* Version Independent ID: e040197a-12cc-281a-d75d-16395b24c46f
* Content: [Microsoft Purview Compliance Manager templates list - Microsoft Purview (compliance)](https://learn.microsoft.com/en-us/microsoft-365/compliance/compliance-manager-templates-list?view=o365-worldwide)
* Content Source: [microsoft-365/compliance/compliance-manager-templates-list.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/compliance/compliance-manager-templates-list.md)
* Service: **o365-seccomp**
* GitHub Login: @chvukosw
* Microsoft Alias: **chvukosw**

msbemba:
@mialavada Thank you for your feedback. We have made changes to the document via PR#https://github.com/MicrosoftDocs/microsoft-365-docs/pull/10851

Once the author approves, the changes will be live.

Thanks
Sri

msbemba:
@yogkumgit  As per Author These updated names don't match what's in the UI for the templates. He has confirmed with the PM today that we should leave the names as-is, not make these changes.

This can be closed.

yogkumgit:
@msbemba Thank you very much for the contribution and sharing this explanation. @mialavada Hope this comment is helpful for you. If you see a documentation update is required, please feel free to open an issue for the same. We proceed here to close it. Thanks for taking out some time to open the issue. Appreciate and encourage you to do the same in future also.

mialavada:
Hi - I work at CIS and am the Product Manager for the CIS Benchmarks. Can I request a reconsideration on this, please? It is an inaccurate representation of our products.

Thank you,
Mia LaVada

PR #10851, opened 2023-01-31
Update compliance-manager-templates-list.md

msbemba:
Made changes to the hyperlink and text
fixes#https://github.com/MicrosoftDocs/microsoft-365-docs/issues/10836


msbemba:

chvukosw:
Need to confer with the PM on this.

chvukosw:
These updated names don't match what's in the UI for the templates. I confirmed with the PM today that we should leave the names as-is, not make these changes.

PR #10931, opened 2023-02-07
Updating service health incident page

seakim-ms:
Updated 2nd table formatting


JohanFreelancer9 (commented):
Copy review - the copy is fine but I am not sure about changing the title. Also, not related but I have noticed that there is a period (.) missing in line 55 between "**Category**" and "The".

seakim-ms:
Thank you. I've added the missing period at the bottom. 

It's important that we change the title as it is more appropriate to what the feature is about.

JohanFreelancer9 (approved):

yogkumgit:
@efrene Please merge this PR. Thanks

PR #10932, opened 2023-02-07
Routing to onprem and external are not allowed

dhairyya:
Updated documentation to prevent escalation
https://o365exchange.visualstudio.com/IP%20Engineering/_workitems/edit/3344969?src=WorkItemMention&src-action=artifact_link


dhairyya:
@chrisda could you please merge this change?

PR #10943, opened 2023-02-08
Update retention.md

kyracatwork:
Adding callout that IM policies are impacted by retention label policies as well.


cabailey:
#sign-off  As discussed

PR #10949, opened 2023-02-09
Removal of duplicate entry for enableFileHashComputation

anoliveiraM:
Removed wrong entry about enableFileHashComputation this doc had two entries regarding enableFileHashComputation, one stating the default value was true(which was not correct as per tests made) and another saying that the default value was false which was the correct value


denisebmsft (approved):
Thank you @anoliveiraM

PR #10977, opened 2023-02-10
Update deploy-manage-removable-storage-group-policy.md

HarveyHuang443:
We need to add the (optional) next to "Enable or Disable Device control" because turning this setting on is optional for Device Control to work. This is to keep it in line with our Intune documentation which also lists it as optional: https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/deploy-manage-removable-storage-intune?view=o365-worldwide#deploy-removable-storage-access-control-by-using-intune-oma-uri


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @HarveyHuangMSFT

PR #10990, opened 2023-02-13
Adding the generic autodiscover URI to the list

brianreidc7:
I had a client who did not use the autodiscover.domain.com value internally as the article here says to add your internal domains - but now that Basic Auth has been switched off for AutoDiscover, you need the autodiscover.contoso.com value added as well or you get a sign-in error when Outlook tries to do AutoDiscover. So you need to include you internal and external domains which includes AutoDiscover, so added something to here to call this out more clearly


JohanFreelancer9 (approved):

yogkumgit:
@kelleyvice-msft Please merge this PR. Thanks

yogkumgit:
@kelleyvice-msft Gentle reminder

PR #11109, opened 2023-02-22
Added the word existing

dhairyya:
Customers are confused and so added the word 'existing'


dhairyya:
@chrisda  could you please approve this for me?

chrisda:
@dhairyya, beware of changing section titles as that can (and in this case did) break links to those sections in other topics.

PR #11128, opened 2023-02-23
Removing beta tags, adding known issues

tdoucett-ms:
@siosulli please review and approve when available.


siosulli:
Closing here and will submit a PR from the private repo.

PR #11136, opened 2023-02-23
Proofread and updated article

dhairyya:
Proofread and updated article


dhairyya:

chrisda:
@dhairyya, again, I beg you, stop making changes to these topics individually. Make the same changes to the 3-4 topics at once so consistency is maintained. And wait until your last batch of changes are published (which would include my edits) so you can see what actually needs to be changed.

PR #11140, opened 2023-02-24
Update m365-lighthouse-requirements.md

scgitdev:
Line 29 - Ensuring clarification that partners could manage customer tenants outside of CSP.


SKjerland:
@scgitdev - Thanks for helping us to improve the article! I reviewed your change with a member of the Lighthouse product team and they requested that we add this as a note instead. We also rewrote it a bit to make sure it's very clear. The updated article should go live this afternoon.

PR #11206, opened 2023-03-02
Update deployment-vdi-microsoft-defender-antivirus.md

agoldenlife:
Instead of using a "cmd" command in the powershell script to expand the security updates; Use the alternative powershell method to invoke the command, better and more flexible.


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @agoldenlife

PR #11255, opened 2023-03-07

kurtsarens:
In passive mode, customer may still see scans running on a schedule. It turned out that these were scheduled in the Windows Scheduler before Defender went into Passive mode. These tasks are not removed or disabled. Removing these tasks is the fastest method as Defender will create these tasks when needed


denisebmsft (approved):
Thanks so much @kurtsarens

PR #11260, opened 2023-03-07
Updated preview information and language.

scorp508:
Included some key information such as.

- This is only for Enterprise Agreement customers
- Customers must still create the initial Mail Enabled User object themselves


scorp508:
I noticed a typo in the PR, it says 'deocrate' instead of 'decorate'

PR #11263, opened 2023-03-07
Update get-started-with-data-lifecycle-management.md

cabailey:
Confirmed by engineering - bug fix rolling out so Scope Manager shouldn't be needed if have Retention Management, so remove from documentation.


cabailey:
#sign-off

PR #11332, opened 2023-03-14
More detail on adding exceptions via GPO

fromalexx:
Currently the instructions for adding exceptions to Folder Access via GPO is missing important information. This caused me a lot of research and time to get these exclusions to work. I added what the columns needs to be set to in order to add an exclusion.


denisebmsft (approved):
Thank you @alexander7567. I appreciate your help.

PR #11409, opened 2023-03-21
Fixed "rolling out" state for AIP Disabled by Default

pefrem:


cabailey:
Needs changing to "Current Channel: Rolling out to 2302+" (so we keep the channel name listed first for consistency)

cabailey:
sign-off, merging

PR #11451, opened 2023-03-23
Update ediscovery-search-and-delete-teams-chat-messages.md

aaronmi:
Removed parentheses and single quotation marks from the section "Purge the chat messages". No need to surround the ID values with parentheses and single quotation marks. Otherwise, The API returns badRequest error.


aaronmi:
Submitted document request to PM.

PR #11531, opened 2023-03-30
Updated scenarios allowed for URL allow

dhairyya:
the left tilde type of entries are not allowed
The left and right tilde type of entries are not allowed


dhairyya:
@chrisda  could you please take a look at this?

PR #11591, opened 2023-04-05
Update api-power-bi.md

edberei:
This documentation does not specify how to access advanced hunting via Power BI with Application context. This document only goes over user context. This is why I'm asking to remove the part that says "build an azure app". It is allowing customers to believe that application context in this scenario is supported and it's not.


JohanFreelancer9 (approved):

yogkumgit:
@Dansimp Please merge this PR. Thanks

PR #11705, opened 2023-04-18
Update microsoft-365-vpn-implement-split-tunnel.md

alexmicrosoft:
We need to replace the old Teams IP CIDR 52.120.0.0/14 with the new one 52.122.0.0/15.


denisebmsft (approved):
Thank you @alexmicrosoft

PR #11710, opened 2023-04-18
Update set-up-multi-factor-authentication.md

### LINKED ISSUE #11581: Turning off Per-user MFA before enabling CAP
Hello, 

We have some concern regarding the note add which request to turn off both Per-user MFA and security defaults before you enable conditional access policy
image

We understand that a customer should not be using both Per-user MFA and conditional access policy that requires MFA at the same time since this will cause a conflict and will contradict with the purpose of using CAP. However, we need to understand why Per-user MFA should be turned off before enabling CAP since some customer are concerned that if Per-user MFA has been disabled before enabling CAP there will be a short gap that will allow users to sign-in without MFA prompt which will raise security risks

Our request is either to remove "Per-user MFA" from the note or add more clarification to the customer so they understand why and what is expected if Per-user MFA was still enabled when CAP is enabled and what risks will this impose 

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: be1e6c9a-9644-6d86-3962-d36bf1efd656
* Version Independent ID: 96142d6c-71de-6ea9-0361-dbec82647b15
* Content: [Set up multifactor authentication for users - Microsoft 365 admin](https://learn.microsoft.com/en-us/microsoft-365/admin/security-and-compliance/set-up-multi-factor-authentication?view=o365-worldwide#use-conditional-access-policies)
* Content Source: [microsoft-365/admin/security-and-compliance/set-up-multi-factor-authentication.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/admin/security-and-compliance/set-up-multi-factor-authentication.md)
* Service: **microsoft-365-business**
* GitHub Login: @KwekuA
* Microsoft Alias: **kwekua**

msbemba:
made changes to the important note that disabling MFA is required after enabling conditional access as it will result in inconsistence user experience

Fixes https://github.com/MicrosoftDocs/microsoft-365-docs/issues/11581


msbemba:
@JohanFreelancer9 PLease copy review

yogkumgit:
@JohanFreelancer9

PR #11718, opened 2023-04-18
Devices currently in troubleshooting mode

cseelhoff-ms:
The KQL for Devices currently in troubleshooting mode is misleading. Either the title should state "Lastest timestamp of when device entered troubleshooting mode" or modify the KQL as proposed


denisebmsft (approved):
Thank you @cseelhoff-ms

PR #11774, opened 2023-04-20

doim-microsoft:
Updating Current Channel descriptions for Disable AIP by default and Louder Label
@cabailey


cabailey:
@doim-microsoft  Looks great, merging to go live.  However, this is only for WXP and not for Outlook that has the same rows.  Is somebody else updating the Outlook table or will you be creating a separate PR for those updates?

PR #11786, opened 2023-04-21
Removing incorrect info about ASR rule default state

sujitnaray:


JohanFreelancer9 (approved):

sujitnaray:
@denisebmsft , @JohanFreelancer9 - Could you submit this change into the final docs? :) Thanks.

v-britweston:
This might be correct, but I need to verify this revision with Engineering. It's possible that it is tied to a specific version / configuration and just needs clarification.

v-britweston:
Approved per Engineering.

PR #11859, opened 2023-04-28
Same tenant channel support is here

dhairyya:
Channel support is finally here for within the same tenant


dhairyya:

PR #11910, opened 2023-05-04
Remove 14 character requirement

cnadler-MSFT:
The guidance in this doc does not align with our more updated recommendations for protecting M365 admins.  M365 admins should be cloud-only, password-less.  In addition, AAD only environments cannot enforce a 14-character password minimum.  I've consulted with the AAD PG on this and happy to discuss in detail as to why this recommendation is not valid.  Thanks!


JohanFreelancer9 (approved):

yogkumgit:
already made this update in the topic so you can removed. Closing

PR #11992, opened 2023-05-11
Updating guidance about end user reporting

dhairyya:
There has been some confusion here with end user reporting for security vs compliance and so updated the documentation to give clarity, Also changed submission to end user report in a sentence to make sure that it is consistent


dhairyya:

PR #11993, opened 2023-05-11
Update microsoft-defender-endpoint-mac.md

YongRhee-MSFT:
Updated entry to include support of M2 CPU.


YongRhee-MSFT:
Need to make changes in a different branch.  Thx.

PR #12088, opened 2023-05-23
Updated SKU information

dhairyya:
MDO for P2 does not support MDE file submission


dhairyya:
@chrisda could you please approve this change for me?

PR #12107, opened 2023-05-25
Update modern-desktop-deployment-and-management-lab.md

cdmm12:


JohanFreelancer9 (approved):

cdmm12:
I received this notice - but not sure what the action item is on my side. Is this set to publish? Or do you I need to commit/approve something?

PR #12165, opened 2023-06-01

samschan-ms:
@KCCross to review and approve. reverted behavior.


KCCross:
Wondering if we should make this an option instead?

PR #12262, opened 2023-06-16
Update anti-spam-policies-configure.md

nidhalferchichi:
Move message to Junk Email folder: is available for high confidence phishing


chrisda:
@nidhalferchichi, true, but it doesn't work if you select it. I'd like to leave it blank to emphasize the fact that it doesn't work. If you can think of a better way to say, "yes, it's available, but don't do it because it doesn't work", I'm all ears.

JohanFreelancer9 (approved):

chrisda:
@nidhalferchichi, there was always a superscript note in the first cell of the 'Move to Junk' row, but I moved the superscript directly into the empty High confidence phishing cell.

So, closing this PR.

PR #12295, opened 2023-06-21
Update activity-reports.md

dotnvo:
In my several talks with Microsoft, we've ran into pretty long delays in reporting data in several areas of teams. This document seems to cover most general reporting in Teams. These delays are seen in Graph, as well as the admin center, which makes sense, because the raw source of the data for these calls are likely the same.

I was informed by support that 48 hours is actually the SLA for this data. However it's common to have a delay of data showing up that is longer than this. Some organizations use this data frequently and it becomes a cause for concern. My suggestion her is to basically add some verbiage that gives Microsoft a little wiggle room and also so IT staff can point to a doc stating that look, it indeed can take longer. Again I've been told that the SLA is 48 hours for teams data to become available, but it's not always the case, particularly around the start of the week, the data seems to be delayed. It takes more than 48 hours often enough for a tenant to get updated info that this document (and likely others) really needs to indicate that it can take more than 48 hours for this to happen. I left it open-ended in my addition, but I welcome Microsoft to modify as needed, I just ask that something is put here to help Microsoft customers who work in IT to able to reference. I've worked with support enough on this to know they don't really have control, and I think the folks responsible for dealing with the data in teams would likely agree with my assessment here.


microsoft-365/admin/activity-reports/activity-reports.md
```diff
@@ -34,7 +34,8 @@ Check out [Microsoft 365 small business help](https://go.microsoft.com/fwlink/?l
 
 You can easily see how people in your business are using Microsoft 365 services. For example, you can identify who is using a service a lot and reaching quotas, or who may not need a Microsoft 365 license at all. Perpetual license model won't be included in the reports. 
   
-Reports are available for the last 7 days, 30 days, 90 days, and 180 days. Data won't exist for all reporting periods right away. The reports become available within 48 hours.
+Reports are available for the last 7 days, 30 days, 90 days, and 180 days. Data won't exist for all reporting periods right away. The reports typically become available within 48 hours, but sometimes take several days
+before becoming available.
```
JohanFreelancer9:
```suggestion
Reports are available for the last 7 days, 30 days, 90 days, and 180 days. Data won't exist for all reporting periods right away. The reports typically become available within 48 hours, but might sometimes take several days
to become available.
```

JohanFreelancer9 (changes requested):
Copy review - @dotnvo, kindly commit the suggested change. Many thanks.

dotnvo:
@JohanFreelancer9 Thanks for your reply. I have implemented your suggestions , should be good to go.

JohanFreelancer9 (approved):

yogkumgit:
@camillepack Please merge this PR. Thanks

PR #12323, opened 2023-06-26
Update find-defender-malware-name.md

YongRhee-MSFT:
Fixing misspelling.


denisebmsft (approved):
Thank you @YongRhee-MSFT Good catch!

PR #12367, opened 2023-07-03
Updated tampering detection titles

j0shbregman:
Updated the detection titles to align with tampering threat analytics article.


denisebmsft (approved):
Thank you so much @j0shbregman

PR #12397, opened 2023-07-06
Update dlp-policy-reference.md

arunt14:
Support for PDF has been removed for password protected predicate on endpoint


yogkumgit:
@arunt14 Thanks for your feedback! New issues and pull requests for this content are closed while we update this content to support upcoming feature improvements.

PR #12486, opened 2023-07-18
troubleshoot asr rules code vs image

ChadSimmons:
Replace adv-hunt-querynew.png with the actual code to allow copy functionality


denisebmsft (approved):
Thank you @ChadSimmons

PR #12680, opened 2023-08-24
Update attack-simulation-training-get-started.md

chrisda:
Added POL as an available region for AST.


richa-b23:
thanks @chrisda

PR #12699, opened 2023-08-29
Fixes typo in attack-surface-reduction-rules-deployment-operationaliz…

Te-k:
Hi,

I noticed a typo in a KQL query in the [Operationalize attack surface reduction (ASR) rules](https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/attack-surface-reduction-rules-deployment-operationalize) page, here is a fix for it.


denisebmsft (approved):
Thank you so much for catching that @Te-k! We appreciate the update.

PR #12714, opened 2023-08-30
Update microsoft-defender-endpoint-linux.md

JesseEsquivel:
Updating docs page to reflect that running FAPolicyD with MDE in active mode is not supported per recent IcM.


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @JesseEsquivel

PR #12804, opened 2023-09-15
Fixed typo in get-machines.md, OData filter parameter $stop should be…

danneforslund:
… $top


denisebmsft (approved):
Thank you @danneforslund

PR #12825, opened 2023-09-20
Adding support for top level domain URL blocking

dhairyya:
Added syntax for top level domain URL blocking


dhairyya:
@chrisda / @MSFTTracyP could you please approve this change and add a reference to what is new inside MDO?

PR #12863, opened 2023-09-26
Update attack-simulation-training-get-started.md

richa-b23:


richa-b23:
@chrisda Attack Sim is now available in the GCC high environment. Updated it in the documentation.

richa-b23:
@chrisda in the below given change it seems like the advanced features are unavailable for both GCC and GCC High environments. Whereas, these advanced features are unavailable only in GCC High environment. 

"As of September 2023, Attack simulation training is available in GCC and GCC High environments, but certain advanced features are not available (for example, payload automation, recommended payloads, the predicted compromised rate)."

chrisda:
@richa-b23, I made the following correction (in bold):

As of September 2023, Attack simulation training is available in Microsoft 365 GCC and GCC High environments, but certain advanced features are not available **in GCC High** (for example, payload automation, recommended payloads, the predicted compromised rate).

Issue #12864, opened 2023-09-26
Add `*.blob.core.windows.net` under OneDrive for Business


kaylangan:
`*.blob.core.windows.net` should also be included under OneDrive for Business since it is used for OneDrive for Business telemetry.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 66bd8666-3566-a901-b25b-eb6baba6264a
* Version Independent ID: e9f6310f-2207-8c81-d31a-3d5a80beb26b
* Content: [Office 365 URLs and IP address ranges - Microsoft 365 Enterprise](https://learn.microsoft.com/en-us/microsoft-365/enterprise/urls-and-ip-address-ranges?view=o365-worldwide)
* Content Source: [microsoft-365/enterprise/urls-and-ip-address-ranges.md](https://github.com/MicrosoftDocs/microsoft-365-docs/blob/public/microsoft-365/enterprise/urls-and-ip-address-ranges.md)
* Service: **microsoft-365-enterprise**
* GitHub Login: @kelleyvice-msft
* Microsoft Alias: **kvice**

msbemba:
@kaylangan Thanks for reaching out to the Microsoft 365 Enterprise content team. Because all Microsoft 365 endpoints data displayed in learn.microsoft.com articles is autogenerated by the Microsoft 365 endpoints web service, the content team cannot resolve issues directly related to this data.
We recommend that you contact the engineering team associated with the endpoint in question.

Hope this helps!

Thanks
Sri

PR #12865, opened 2023-09-26
Add `*.blob.core.windows.net` to OneDrive for Business

kaylangan:
See #12864 

Not sure if this is the right place to suggest updates given the header says that this file is automatically generated.


scotvidican:
@kaylangan - you're correct in that it's autogenerated from a db, so we can't update it manually. I'll send you mail with information on how to request changes.

PR #12884, opened 2023-09-28
Update ios-install.md

kerodri:
needs to be specific on what license is needed


Dansimp:
Will review and update in a separate PR

PR #12907, opened 2023-10-05
E bpf doc update

meghapriyams:


microsoft-365/security/defender-endpoint/linux-support-ebpf.md
srkoripe, line 1: Just wondering if we can remove this statement.
> Customers need to update to the above-mentioned supported versions to experience the feature

PR #13005, opened 2023-10-23
ASR rules: Provide minimal info on "Block Webshell creatieon for Serv…

matsimon:
…ers"

Partially addresses MicrosoftDocs/microsoft-365-docs-pr#23612 but lots of parts cannot be documented without internal knowledge from Microsoft.

If you search for this ASR rule, it has already been out without any documentation for quite some time.
Please fill in the missing gaps if you can, this at least will give you a hint on what GUID is being used.


denisebmsft (approved):
Thank you @matsimon

PR #13009, opened 2023-10-23
Graph and user impersonation are same

dhairyya:
Customers are getting confused whether graph and user impersonation are one and the same thing. So making a change to clear it out.


dhairyya:
@chrisda could you please approve this?

PR #13011, opened 2023-10-23
Added AIR outcome for third-party

dhairyya:
Customers need to be educated the advantage of tying up third party report on user reported settings. The biggest advantage is AIR and so calling it out


dhairyya:
@chrisda could you please approve this?

PR #13031, opened 2023-10-26
Update printer-protection-overview.md

PiotrWadolowski:
Update printer-protection-overview.md - moved "print 64" to correct location in access mask


denisebmsft (approved):
Thank you @PiotrWadolowski

PR #13065, opened 2023-11-01
Updated sender blocking action

dhairyya:
Sender block leads to quarantine instead of SCL:9


dhairyya:
@Dansimp could you please merge this PR?

PR #13075, opened 2023-11-02
Adding Italy multi-geo availability - Update microsoft-365-multi-geo-…

r2k8:
…locations.md

Italy multi-geo is now available.
Arranged alphabet order, swapped Poland and Qatar


r2k8:
Adding Italy multi-geo to the list

PR #13109, opened 2023-11-07
Update cross-tenant-onedrive-migration.md

wonderboy703:
This needs to be clear, the term created could be convoluted. It should refer to pre-provisioned where someone accesses OneDrive or someone tickles the data source thus creating the OneDrive URL.


wonderboy703 (commented):
I might of messed the link up, it appeared to be correct. I was trying to define pre-provision with this article. 

https://learn.microsoft.com/en-us/sharepoint/pre-provision-accounts

PR #13116, opened 2023-11-08
Update attack-surface-reduction-rules-deployment-operationalize.md

Eagle-Admin:
Misspelling on line 77. Changed 'i' to the correct 'j' in "ExtractJson".


denisebmsft (approved):
Thank you so much @Eagle-Admin

PR #13183, opened 2023-11-17
Update mde-plugin-wsl.md

michaelmsonne:
Fix the status "A positional parameter cannot be found that accepts argument 'Files\Microsoft'" when copy the command and input it in the console raw - as space not counted :)


denisebmsft (approved):
Thank you @michaelmsonne

PR #13235, opened 2023-11-28
Malicious sites are phish

dhairyya:
Phish were incorrectly defined as spam


dhairyya:
@chrisda could you please merge this change?
Thanks in advance

PR #13245, opened 2023-11-30
TABL can support more entries

dhairyya:
Updated whats new with TABL support of more entries in November 2023


dhairyya:

PR #13259, opened 2023-12-01
Confusion about similar URL allow

dhairyya:
Rectified the earlier example which was shared here


dhairyya:
@chrisda could you please approve this for me

PR #13260, opened 2023-12-01
confusion about similar URL allow

dhairyya:
Replaced earlier example which was creating confusion


dhairyya:

PR #13280, opened 2023-12-06
Update whats-new-in-microsoft-defender-endpoint.md

noamhadash:
adding the release of User Contain in October


noamhadash:
fixed broken link

PR #13316, opened 2023-12-15
Fix typos

Blake-Madden:
Misspellings, double words


denisebmsft (approved):
Thank you so much @Blake-Madden

PR #13327, opened 2024-01-17
Simplified secops mailflow bypassing

dhairyya:
Removed guidance around antimalware as advanced delivery skips malware filtering as well


dhairyya:
@chrisda could you please merge these changes

PR #13332, opened 2024-01-18
Used allow incorrectly instead of block

dhairyya:
Fixed a typo in the documentation


dhairyya:

PR #13333, opened 2024-01-19
Update m365-dr-workload-mdo-p1.md - added missing header

jbrianmoss:
The document was missing a header for Microsoft Defender for Office P1, and went straight to the Overview header.


kelleyvice-msft:
This change would break the article structure convention shared with related articles.

PR #13334, opened 2024-01-22
Sender block are HCP

dhairyya:
Customer call to call out specifically what is the type of phishing sender block lead to


dhairyya:

PR #13343, opened 2024-01-24
Update backup-limitations.md

brguss:
Added note about EXO multi-geo support in the limitations section.


brguss:
Added EXO multi-geo support in limitations section

PR #13346, opened 2024-01-24
Update manage-data-gcc.md

alexsfaulkner:
Added line about comments not being included in ediscovery.


v-jmathew:
As per the response from submitter, they no longer work on Whiteboard so unfortunately, they cannot answer this.

PR #13347, opened 2024-01-25
Updated 90 to 30

dhairyya:
For emails it is not 90 days from system date. It is 30 days from system date


dhairyya:

PR #13358, opened 2024-02-01
Update manage-rbac.md

marinakidron:
to highlight the B2B disclaimer will be specific for the MDO piece


marinakidron:
@chrisda additional minor update, to highlight that it only refers to MDO

PR #13367, opened 2024-02-07
Add a list of no-reply sender addresses

jpawlowski:
This would add a (not complete) list of sender addresses that are currently used by Microsoft services.

This list might be useful for people to know when dealing with delivery issues. It might also help to even prevent delivery issues in the first place as by knowing the sender address in advance, proper handling can be ensured by the admin team until the service is supported using the centralized setting, or a client might not want to use the centralized setting for whatever reason.


v-jmathew:

jpawlowski:
> Closing the PR as GitHub account of the submitter not found.

But I'm here. @jpawlowski, it exists.

PR #13371, opened 2024-02-07
Removed 'second opinion' to avoid confusion

dhairyya:
Removed terminology second opinion which was causing confusion to customers.


dhairyya:

PR #13381, opened 2024-02-13

kurtsarens:
Updating list adding exclusions


denisebmsft (approved):
Thank you @kurtsarens

PR #13382, opened 2024-02-13
Update microsoft-defender-endpoint-mac.md remove Big Sur as supported version

marckerr:
If the three most recent major releases are the only one supported then the listing to Big Sur needs removed.


Dansimp:
@YongRhee-MSFT , can you confirm this change? If so, I will merge. Thx!

YongRhee-MSFT:
Hi Dan,

Thx for following up.

Confirming, Apple dropped support for BigSur at the end of Nov '23.
***@***.***
Thx,
Yong

PR #13413, opened 2024-03-01
Update manage-self-service-signup-subscriptions.md

JonathanBrockhausen:
The displayname of the hyperlink does not match the actual URL and also the PowerShell Parameter is wrong under the new cmdlet


JonathanBrockhausen:
Also the linked URL refers to the same cmdlet and is also outdated here. Should also be refreshed but requires a different commit.

PR #13418, opened 2024-03-05
Fix oAuthUri in exposed-apis-create-app-webapp.md

hancockl:
The PowerShell code won't get the proper token using the oAuthUri as written. Needs to reference the v2.0/token endpoint


denisebmsft (approved):
Thank you @hancockl

PR #13432, opened 2024-03-12
Update custom-detection-rules.md

Yanivik:
add a comment that in NRT the query do not include comments


denisebmsft:
Macky already updated.

PR #13434, opened 2024-03-13
Remove domain sharing information, add notes about quota management for moves.

ropow:
1. Remove reference to domain sharing as it is currently on hold.
2. Update document to give best practices for managing mailboxes close to, or over quota, prior to beginning migration.


ropow:
@kelleyvice-msft This can be checked in, I have no further changes to make.

PR #13441, opened 2024-03-15
Update azure-ad-setup-guides.md

v-fentonmacy:
Updated content for Configure multifactor authentication (MFA)


v-fentonmacy:
@kwekuako   updated content for Configure multifactor auth

PR #13445, opened 2024-03-19
Update esignature-setup.md

aoifemcd1:
Changed signers to recipients for the External signers section and changed some of the ordering to be more logical


chuckedmonson:
Changes incorporated.

PR #13473, opened 2024-03-26
Update about-shared-mailboxes.md

Smartboss88:
Update doc per Branden Lee's comment that shared mailboxes don't need license if org wide retention policy is applied.


Smartboss88:
Any update ?

v-jmathew:

PR #13475, opened 2024-03-27
Update manage-sys-extensions-using-jamf.md

robatjamf:
Issues with how Jamf was written. Updated

It is not JamF or JAMF

Just Jamf


robatjamf:
@microsoft-github-policy-service agree

PR #13485, opened 2024-04-03
Update attack-simulations.md

noamhadash:
update that evaluation lab was deprecated


denisebmsft (approved):
Thank you @noamhadash

PR #13488, opened 2024-04-04
Added default wsl version to use for mde

abhiksen:
Distributions running on WSL1 are not supported to run MDE for WSL plugin, and hence users need to disable WSL1 or specifically use WSL2 distributions to run MDE for WSL plugin.


denisebmsft (approved):
Thank you @abhiksen

PR #13500, opened 2024-04-08
Update whats-new-in-microsoft-defender-endpoint.md

noamhadash:
adding that User Contain has moved to GA in January


denisebmsft (approved):
Thank you @noamhadash

PR #13514, opened 2024-04-11
Update select-domain-to-use-for-email-from-microsoft-365-products.md

Stealth7777:
Updated to highlight that EXO admin management is needed if a mailbox is configured.


v-jmathew:
As confirmed by the submitter, the change is still relevant.

PR #13521, opened 2024-04-12
Update run-analyzer-macos-linux.md

liam-x:
Now client analyzer on macOS have single executable support both Intel and ARM mac.


denisebmsft (approved):
Thank you @liam-x

PR #13538, opened 2024-05-03
Update network-requests-in-office-2016-for-mac.md

larspassic:
The URL officecdnmac.microsoft.com was announced in 2021 release of Office for Mac AutoUpdate, however the URL was never added to documentation.

https://learn.microsoft.com/en-us/officeupdates/release-history-microsoft-autoupdate#october-2021-release

I was not sure which type of URL it was (static or semi static) so please double check this for accuracy!


larspassic:
@kelleyvice-msft can you review this? Thank you!

PR #13546, opened 2024-05-15
Updated ABT Activation and FAQ Sections

ByBrianGoodman:
1. Added additional detail regarding which roles are required for initial activation of Idle Session Timeout.
2. Added additional note to FAQ for Azure Portal timeout inheritance from M365.


ByBrianGoodman:
@kwekuako - Checking on status of this PR. Thanks!

PR #13601, opened 2024-07-23
Update manage-self-service-purchases-users.md

mihart:
removed mentions of trials since this article is about purchases.  We link from PBI and we noticed that including trials isn't relevant here.


mihart:
#sign-off

cmcatee-MSFT:
PR is in public repo, which can't be merged with the private repo. Per PM, article applies to trials, as well. Updates made accordingly.

PR #13608, opened 2024-07-31
Update custom-domain-support.md

alexlmit:
Added clarification in the document so it's clearer that the custom domain MUST be configured as an Accepted Domain in the tenant's domains list. It's not explicitly clear to end users right now.


v-jmathew:

PR #13616, opened 2024-08-12
Update modern-desktop-deployment-and-management-lab.md

cdmm12:


aczechowski (approved):
Thanks @cdmm12 - I also bumped the ms.date metadata since it had been about 6 months, so just noting the article freshness. :)

PR #13630, opened 2024-09-06
Update manage-feedback-ms-org.md

karanmalik-sudo:
Removed redundancy in the table with both Microsoft Whiteboard and "Whiteboard" being present as two separate entries.


karanmalik-sudo:
@kwekuako could you please approve this PR?

PR #13639, opened 2024-09-17
Update turn-off-directory-synchronization.md

frenzied-state:
Updated the command Update-MgBetaOrganization to Update-MgOrganization. The beta version of the command no longer works. I've tested this in my dev tenant and in a production environment. I've linked the Learn article for Update-MgOrganization below. I'd also suggest removing the command to install the MgGraph beta, but I haven't tested that.

https://learn.microsoft.com/en-us/microsoft-365/enterprise/turn-off-directory-synchronization?view=o365-worldwide


kelleyvice-msft:
No file updated, closing.

PR #13676, opened 2024-11-25
Update setup-guides-for-microsoft-365.md

mitackie:
Microsoft Places ADG & Setup entry point


kelleyvice-msft:
@mitackie, I just found this PR assigned to me on 11/25/2024. For some reason I wasn't notified. Can you confirm that this update is still valid and needs to be published? Thanks.

PR #13677, opened 2024-11-27
docs: manage-guest-access-in-groups.md - Fixed typo

mitchplze:
I can't see any reason why these asterisks should be there.


v-mathavale:
Changes are already there in the private repo, hence closing.

PR #13679, opened 2024-12-05

realPrix:
The old code only output a limited paged number of elements if you have a lot of groups. My commit fixes this problem by filtering the output before.


realPrix (commented):
I did this by accident, please ignore

realPrix:
@microsoft-github-policy-service agree

PR #13694, opened 2024-12-23
Update create-secure-guest-sharing-environment.md

WilliamLeggatt:
Added Governance ID license comment to Guest Access review section, as this capability requires it.


v-mathavale:
Change is technical and can't trace the autor.

PR #13704, opened 2025-01-09
Update password administrator to user administrator

DotMindy:
Per discussion with PG and also lab test, the "password administrator" role is unable to update the user's password policy while user administrator can.

Refer to the ICM
https://portal.microsofticm.com/imp/v5/incidents/details/577573272/summary?tmpl=UW162p
image


v-mathavale:
Substantial change and author can't be found.

PR #13705, opened 2025-01-09
Update admin note in manage-office-scripts-settings.md

alison-mk:


SKjerland:
@alison-mk - I made a minor edit to the sentence to follow our security recommendations around the use of "Global Admin" in our help articles. Also, I noticed you updated the **ms.date** metadata field. This field should only be updated if you reviewed the entire article and confirmed that the content is correct and up to date. Did you do that? I just want to make sure the article is actually up to date and doesn't require a freshness review from a SME. Thanks!

alison-mk:
@SKjerland Thanks for the feedback and merging the PR! Based on your criteria, I shouldn't have updated the ms.date field. My apologies! Should I change it back?

PR #13708, opened 2025-01-10
Update TOC.yml with Viva Engage

danthemango:
Rename Yammer to Viva Engage in TOC
to reflect updated branding

This change is related to commit d2ac96c139290328ee8bd54706d4a6ff7ac6fbd8 which failed to update the TOC.


danthemango:
I'm unsure if we should also update the links too, they currently redirect to https://learn.microsoft.com/en-us/viva/engage/overview but based on the context (deploy, manage) I believe urls should be update to something like https://learn.microsoft.com/en-us/viva/engage/setup and https://learn.microsoft.com/en-us/viva/engage/eac-overview

v-mathavale:
Substantial change and can't trace the author.

PR #13712, opened 2025-01-17
Update microsoft-365-copilot-page.md

elvafdz:
Added more settings and updated the learn more urls for that new sections


elvafdz:
@camillepack I tried to do the changes above and create a PR, can you help with the updates I believe on the URLs I need input, thanks

PR #13730, opened 2025-02-04
Update select-domain-to-use-for-email-from-microsoft-365-products.md

Stealth7777:
Updated to explicitly call out that the tenants own domain would be used as support ask this question even though it is called out multiple times


v-mathavale:
Contacted the author on teams to find out if the changes are still relevant. Waiting for his reply.

v-mathavale:
No reply from author hence closing.

PR #13738, opened 2025-02-20
Adding License Eligibility from Product terms

TiboJFR:
This page could be misleading without the product terms as one can thinks that Frontline workers could use Office Web with a primary device not shared and with a screen bigger than 10.9''


TiboJFR:
Hey. Not sure where to put it in the page but clearly, without those licensing terms, this page does not seem explicit enough :(

v-mathavale:
Technical change and author not found.

PR #13759, opened 2025-03-19

snaggo:
Added -All parameter to Get-MgBetaGroup in line 29 to prevent the script from only searching the first 100 groups. In large environments, this script won't work otherwise because it can't find the specified Group so GroupCreationAllowedGroupId will remain empty and nobody can create groups.


v-mathavale:
changes are made in the private repo.

PR #13762, opened 2025-03-20
msappproxy for entra id join with SSO is missing

Wopienkaatwork:
We encountered issues with Entra ID Domain Join with SSO and figured out that the domain *.msappproxy.net was blocked by the firewall/proxy because it was missing in the list. It is mentioned here https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/how-to-connect-sso-quick-start but not in the summary page. I would suggest to put this domain on the summary page as well.


v-mathavale:
This PR has been open for an extended period of time and has conflicts. If these changes are still relevant, please re-submit them to [microsoft-365-docs-pr](https://github.com/MicrosoftDocs/microsoft-365-docs-pr).

PR #13810, opened 2025-04-18
Update backup-restore-data.md

esung22:
A few minor updates


esung22:
@chuckedmonson Can you please help me with a relative link to the Recoverable Items document?  The absolute link is triggering a warning.  I don't have access to the EXO docs, I guess it has different permissions than the learning center docs.

PR #13813, opened 2025-04-22
Update understand-proposal-workflow.md

titung-MSFT:
Updated to include accounts payable contact. Removing your information as that's not accurate.


cmcatee-MSFT:
Resumitted in private repo

PR #13814, opened 2025-04-22
Adding images for demonstration

minatoruan:
Adding images for deployment at scale with flexible membership document.


v-mathavale:
All changes implemented in private repo.

PR #13818, opened 2025-04-24

SixPraxis:
Added -All switch to Get-MgBetaGroup cmdlet to ensure success in tenants with more groups than the default returned sample size.


v-mathavale:
changes are implemented in private repo.

PR #13827, opened 2025-05-08
Add new policy "Allow users to submit limited feedback to Microsoft"

bajdigital:
There is a new policy "Allow users to submit limited feedback to Microsoft" when "Allow users to submit feedback to Microsoft" policy is disabled.


v-mathavale:
Technical change and author not found.

PR #13833, opened 2025-05-14
Update archive-faq.md

DanielRyanTech:
File-level granular archiving support was pushed back to 2026, but it has been added to the Roadmap for tracking. Updated the FAQ entry to have the correct year, and added a link to the direct entry on the Microsoft 365 Roadmap


DanielRyanTech (commented):
That makes sense, thanks for catching that part of the link!

PR #13842, opened 2025-05-21
Updating Idle Session Timeout handling for Microsoft Fabric

ByBrianGoodman:
Updated FAQs to cover long-running executions present within Microsoft Fabric to bring additional clarity with ABT.


ByBrianGoodman (commented):
Fixed verbiage.

PR #13858, opened 2025-05-28
Update about-admin-roles.md

bharto:
Migration admin role now allows file share migrations as well.


v-mathavale:
technical change and unable to trace author.

PR #13859, opened 2025-05-29
Update OneDrive/SharePoint maximum path character limit

MeckenS:
This should be 247-character limit. Please try creating a file or folder with a path limit of more than 247 characters on your own Windows computer with OneDrive KFM enabled to see this behavior. I can submit a video of this behavior if needed.

Additionally, the sentence could be more descriptive to include maximum path length since that is what is being affected.


v-mathavale:
closing as can't trace author and change is technical.

PR #13860, opened 2025-05-29
Update microsoft-365/whiteboard/manage-sharing-organizations.md

akshaygautam:
The following requirements for setting up Microsoft Teams Room device was missing in doc. Users also need to enable SPO level setting in their tenant to access Whiteboard on devices


akshaygautam:
#sign-off

PR #13882, opened 2025-07-03

leviu-ofek:
Added documentational to link to ability to turn off 365 group creation completely via a GUI Entra ID setting. Not sure why this article never mentioned this ability and it can be done simply via GUI.
[https://learn.microsoft.com/en-us/entra/identity/users/groups-self-service-management#group-settings](https://learn.microsoft.com/en-us/entra/identity/users/groups-self-service-management#group-settings)


v-mathavale:
changes are technical and and the author can't be found.

leviu-ofek:
Author of what? this .md?

